PR #57, opened 2026-02-23
fix(web): allow build without Supabase env vars

howarewoo:
## Summary
- Fix `next build` crash when `NEXT_PUBLIC_SUPABASE_URL` is unset (e.g. CI without env secrets)
- Static pages like `/_not-found` go through the root layout → `<Providers>` → `createBrowserSupabase()`, which threw when env vars were missing

## Changes
- `apps/web/lib/supabase.ts`: Only throw the missing-env-var error at runtime in the browser (`typeof window !== "undefined"`); during SSR/build prerendering, return empty strings so the Supabase client is created but never makes real requests

## Test Plan
- [ ] `pnpm --filter web build` succeeds without `NEXT_PUBLIC_SUPABASE_URL` set
- [ ] `pnpm --filter web dev` still throws if env vars are missing in the browser
- [ ] CI build passes

---

### DIFF
--- a/apps/web/lib/supabase.ts
+++ b/apps/web/lib/supabase.ts
@@ -3,10 +3,17 @@ import { createSSRBrowserClient } from "@infrastructure/supabase/browser-ssr";
 import { createSSRServerClient } from "@infrastructure/supabase/server-ssr";
 
 function getSupabaseConfig() {
-  const url = process.env.NEXT_PUBLIC_SUPABASE_URL;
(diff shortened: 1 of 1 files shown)


howarewoo:
* **#57** <a href="https://app.graphite.com/github/pr/howarewoo/monorepo-template/57?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/howarewoo/monorepo-template/57?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `staging`


<h2></h2>

PR #148, opened 2026-06-01
feat(woostack-review): nest repo config under "review" key

howarewoo:
## What

Review settings in `.woostack/config.json` now nest under a top-level `review` object so the file can hold sibling config namespaces for other woostack tools (build, bootstrap) without collision. Keys outside `review` are ignored by the review loader.

```json
{ "review": { "severity_floor": "medium", "angles": { "skip": ["seo"] } } }
```

## How

- `scripts/load-config.sh` selects the `review` block and still emits a **flat** canonical `$OUTDIR/config.json`. Single change point — every downstream stage (`prefetch`, `detect-angles`, `chunk-diff`, `intersect-findings`, `metrics-fold`, `reusable-review.yml`) reads `.severity_floor`/`.angles`/etc. unchanged.
- **Transition:** review keys at the top level (pre-nesting layout) are still accepted but emit a deprecation `::warning`. Remove the legacy path once consumers migrate.
- Docs (`SKILL.md`) and the `reusable-review.yml` comment updated to the nested form.

## Tested

Parser paths verified: nested, sibling-namespace-ignored, legacy-warn, `review` not-object error, unknown-review-key error, unknown-top-level error, empty/missing-file defaults, full-schema round-trip → flat output confirmed.

### DIFF
--- a/.github/workflows/reusable-review.yml
+++ b/.github/workflows/reusable-review.yml
@@ -179,8 +179,8 @@ jobs:
       # findings.prosecutor.json (inclusive bias), defender pass writes
       # findings.defender.json (exclusive bias) + runs intersect-findings.sh
       # to produce the final findings.json, then posts the review.
-      # `disable_adversarial: true` in .woostack/config.json short-circuits the
-      # prosecutor step (the intersect script then falls back to defender-only).
+      # `review.disable_adversarial: true` in .woostack/config.json short-circuits
+      # the prosecutor step (the intersect script then falls back to defender-only).
       - name: Read disable_adversarial flag
         id: adv
         shell: bash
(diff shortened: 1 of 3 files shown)


howarewoo (commented):
## ✅ APPROVED — woostack-review

**Angles run:** bugs, security, conventions, aeo, infra, docs
**Findings:** 0 (after adversarial validation @ `severity_floor: high`)

No blocking or HIGH-severity issues. The `load-config.sh` nested-`review` resolution, legacy fallback, unknown-key validation, and flat canonical output are correct; security surface is clean (trusted self-owned config, stdlib `json.loads`, no eval/shell path).

<details><summary>Sub-floor notes (MEDIUM/LOW — dropped by the default <code>high</code> floor, not posted as blocking)</summary>

Two genuine but non-blocking doc nits in `SKILL.md` were surfaced and filtered by the severity floor:
- The `authors_skip` opt-out example still shows the deprecated top-level `"authors_skip": []` form.
- The "keys outside `review` are ignored" sentence sits next to the "unknown keys → error" claim; the error path applies to unknown keys *inside* `review`, not sibling namespaces — worth disambiguating.

</details>

> ℹ️ Event downgraded to **COMMENT**: GitHub does not allow the PR author to formally APPROVE their own PR. The verdict above is the real result.

PR #149, opened 2026-06-01
docs: expand README with how-it-works, quickstart, concepts, CI review

howarewoo:
## What

Turns the README from a marketing-style overview into a usage guide, while keeping it lean — each "how it works" section links to the relevant `SKILL.md` rather than restating it (honors the repo's cross-link-don't-duplicate rule).

## Changes

- **Section nav** at the top of the file.
- **How it works** — per-command flow for bootstrap / build / review / address-comments, each with a link to its `SKILL.md`.
- **Quickstart** — concrete install → bootstrap → build → review → address sequence.
- **Concepts** — `.woostack/` artifact layout, the `main`/`staging`/`feature` branching model, and the `detect → fan-out → merge → skeptical validator → post` review flow (prosecutor + defender passes). Folds in the old loose `woo-review` deprecation line.
- **Cloud / CI review** — documents the `action.yml` + `reusable-review.yml` delivery with the consumer workflow snippet (copied verbatim from the review SKILL).
- Default-stack table now notes the stack is confirmed per-project, not mandated.

## Verification

All 15 file links and both section anchors (`#architecture`, `#companion-github-action`) verified to resolve. No application code, build configs, or self-CI added — skill/doc assets only, per AGENTS.md Mode A.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #151, opened 2026-06-01
fix(woostack-review): section-aware prefetch diff cap (#150)

### LINKED ISSUE #150: woostack-review: prefetch 300KB diff cap is order-blind, silently drops high-value files
## Summary

`prefetch.sh` truncates `diff.txt` at a hard 300KB byte cap. The truncation is order-blind: it keeps the first 300KB of the diff verbatim and drops the rest. When a large-but-low-value prefix (mass file deletions, generated files, lockfiles) sorts ahead of the substantive changes, the real review surface gets cut off and the swarm reviews nothing meaningful — while reporting a clean pass.

## Where

`skills/woostack-review/scripts/prefetch.sh`, ~line 380:

```bash
if [ "$DIFF_BYTES" -gt 300000 ]; then
  {
    head -c 300000 "$OUTDIR/diff.txt"
    printf '\n[DIFF TRUNCATED AT 300KB]\n'
  } > "$OUTDIR/diff.txt.capped"
  mv "$OUTDIR/diff.txt.capped" "$OUTDIR/diff.txt"
fi
```

No env override; `head -c` cuts mid-stream regardless of file boundaries or review value.

## Repro (observed on PR #149)

- PR #149: docs/config change (README, AGENTS.md, four SKILL.md, skills-lock.json, install.sh) bundled with bulk deletion of vendored `.agents/skills/**` and `.claude/skills/**`.
- Full `gh pr diff 149`: 76 file sections, ~393KB.
- After the 300KB cap: `diff.txt` held only 46 sections — **all** `.agents/`/`.claude/` deletions (they sort alphabetically first). Every substantive root file (README.md, AGENTS.md, skills-lock.json, the SKILL.md edits) was dropped.
- A swarm run on that `diff.txt` would see only vendored-skill deletions → zero real findings → false "clean".

## Impact

- Silent under-review. The truncation sentinel is appended to the diff body, but nothing fails the run or warns the orchestrator that the *high-value* files were the ones cut.
- Worst case is exactly the common case: a refactor/cleanup PR that deletes a tree (sorts early) plus edits a few key files (sort late).

## Proposed fixes (pick one or combine)

1. **Per-file budgeting** instead of a global byte cut: iterate file sections, include whole sections until the budget is hit, never split a section. Emit a `::warning::` listing which files were dropped.
2. **Order by review value before capping**: sort file sections so likely-low-value sections (pure deletions, lockfiles, generated/`*.md` vendored trees, paths matching `ignore`) go last, so the cap eats those first.
3. **Drop pure-deletion sections from the review diff entirely** (a deletion has no RIGHT side to anchor comments on anyway — `resolve-diff-line.sh` returns `null` for them), or at least deprioritize them.
4. **Raise/parameterize the cap** via env (e.g. `WOO_REVIEW_DIFF_CAP_BYTES`) as a stopgap.
5. **Loud signal**: when truncation drops any file present in `meta.json` but absent from the capped `diff.txt`, emit a `::warning::` and record the dropped paths in an artifact.

(1) + (5) together fixes the correctness and the silence. (3) is a cheap high-value win on its own since deletion hunks can't carry inline comments.

## Workaround used for #149

Rebuilt `diff.txt` from full `gh pr diff 149` minus `.agents/`/`.claude/` deletions (28KB, 9 files), re-ran `detect-angles.sh`, dropped the stale chunk artifacts, then ran the swarm.

howarewoo:
Fixes #150.

## Problem

`prefetch.sh` capped `diff.txt` at a hard 300KB via `head -c 300000` — an order-blind byte cut. When a large low-value prefix (mass file deletions, lockfiles, generated trees) sorts ahead of the substantive changes, the cut drops the real review surface. The swarm then reviews only the low-value prefix and reports clean — a silent under-review. Observed on #149, where `.agents/`/`.claude/` deletions (alphabetically first) consumed the budget and every authored file (README, AGENTS, SKILL edits) was dropped.

## Fix

Section-aware cap:

- Split the diff into whole `diff --git` file sections.
- Rank by review value: tier 0 = sections that add lines (real RIGHT-side content to anchor comments on); tier 1 = modified, no additions; tier 2 = pure file deletions + lockfiles/generated (deletions have no RIGHT side, so `resolve-diff-line.sh` returns `null` for them anyway).
- Greedily keep whole sections until the budget is hit, never splitting one. The single highest-value section is kept even if it alone exceeds the budget (one oversized section beats reviewing nothing).
- Kept sections are re-emitted in original diff order.

Plus:
- **Never silent**: dropped paths recorded to `$OUTDIR/diff-dropped.txt` and surfaced as a `::warning::`.
- **Parameterized**: cap overridable via `WOO_REVIEW_DIFF_CAP_BYTES` (default 300KB).
- Under the threshold the diff is left byte-for-byte untouched (no behavior change for normal PRs).

## Verification

- Synthetic diff (big pure-deletion file + lockfile + small real edit, cap forced low): the tier-0 edit is kept with its additions intact, the deletion dropped + warned + recorded, lockfile kept only because budget remained.
- `bash -n` clean.
- Documented the new artifact + env var in `SKILL.md` troubleshooting.


skills/woostack-review/scripts/prefetch.sh
```diff
@@ -375,14 +375,120 @@ if [ -z "$INCREMENTAL_USED" ] && [ "$LOC_CHANGED" -lt 10 ]; then
...
+# Stable sort by tier, preserving original order within a tier.
+ranked = sorted(sections, key=lambda s: (s["tier"], s["orig"]))
+
+budget = cap - len("".join(preamble).encode("utf-8"))
+kept_idx, dropped = set(), []
+used = 0
+for sec in ranked:
+    # Always keep the single highest-value section even if it alone exceeds the
+    # budget — one oversized section beats reviewing nothing. Never split it.
+    if not kept_idx or used + sec["bytes"] <= budget:
```
howarewoo: **Oversized forced-keep corrupts budget for remaining sections**

When the highest-value section (first in `ranked`) exceeds the cap on its own, the greedy loop adds its full byte count to `used` (line 454). Every subsequent section then fails the `used + sec["bytes"] <= budget` check and is dropped — even tiny high-value sections that would easily fit within the cap independently. The output file also ends up larger than the cap (e.g. a 400KB section with a 300KB cap produces a 400KB output). The PR description promises 'one oversized section beats reviewing nothing', but the implementation silently discards all other sections instead of continuing to fill the remaining budget.

Fix: Track the forced keep separately so it does not count against the budget for subsequent sections. Replace the single `if` with: `if not kept_idx: kept_idx.add(sec["orig"]); used += min(sec["bytes"], budget)` (cap the forced keep's contribution to `budget` so the loop can still add small sections), or introduce a separate `forced_overflow` flag and keep `used` counting only budget-checked sections.

howarewoo (commented):

Reviewed the section-aware diff cap (5 angles: bugs, security, conventions, aeo, docs). One blocking correctness bug in the greedy keep loop: an oversized forced-keep section poisons the byte budget so every later section is dropped. Security, conventions, and the docs accuracy all checked out.

---
**Status: CHANGES REQUESTED** — 1 blocking finding(s) (1 HIGH, 0 MEDIUM, 0 LOW) + 0 non-blocking. See inline comments.
*Audited by woostack-review · Host: Claude Code · Provider: anthropic · Model: claude-opus-4-8*



skills/woostack-review/scripts/prefetch.sh
howarewoo: Fixed in 150671a. The oversized forced-keep is now an allowed one-time overflow that doesn't consume the budget: only bytes that actually fit are added to `used`, so the remaining sections still fill up to the cap instead of all being dropped. Verified with a 166KB top section under a 100KB cap — the tiny following section is now kept (old logic dropped it).

Note: the suggested `used += min(bytes, budget)` was a no-op for this case (it sets `used == budget`, which still starves every later section); the fix instead skips counting the overflow entirely.

howarewoo (commented):

Re-review of `150671a` after the diff-cap fix. The prior HIGH blocking bug (oversized forced-keep starving the budget) is resolved: the forced keep is now a free one-time overflow, so subsequent sections still fill the cap. Bugs angle independently re-verified the loop — no remaining or new correctness issue. Security, conventions, aeo, docs all clean.

---
**Status: APPROVED** — No validated findings.
*Audited by woostack-review · Host: Claude Code · Provider: anthropic · Model: claude-opus-4-8*

PR #172, opened 2026-06-02
feat(review): per-angle cross-angle overlap metric (tokens deferred)

howarewoo:
## What

Adds a per-angle **cross-angle overlap** signal to `woostack-review`'s opt-in metrics, so redundant review angles (those whose findings other angles also raise) are visible per run and across runs.

This is **increment 1 of 2**. The second requested signal — per-angle **token cost** — is deferred to a follow-up (see Deferred below).

## Why

The existing opt-in metrics (`findings.metrics.json` + rolling `.woostack/metrics.json`) record signal/noise per angle but cannot show **redundancy**: when several angles independently raise the same issue, that duplicated effort is invisible. An angle whose findings are nearly always also found by another angle is a candidate to drop.

## How

- **`intersect-findings.sh` → `emit_angle_metrics`**: cluster raw findings by `(file, line, title_stem)` — the merge-findings dedup key minus `angle` — and emit per angle:
  - `overlap_with`: `{other_angle: count}` (total pairwise co-occurrence on the **raw** pre-validation set)
  - `overlap_total`: sum of that map
  - **Unanchored findings** (no `file`/positive-int `line`) are excluded — no phantom clusters.
  - An angle never counts overlap with itself (within-angle dupes already collapsed by `merge-findings.sh`).
  - Per-run doc `schema_version` → 2.
- **`metrics-fold.sh`**: `SCHEMA_VERSION` → 2; accumulate `overlap_total` (sum) and `overlap_with` (merge-add) into the rolling aggregate. A stale v1 aggregate is backed up to `.bak` and reseeded via the existing version-mismatch path (per-clone, gitignored data — acceptable).
- **`SKILL.md`**: artifact-table row + `metrics` config note document the new fields and schema v2.

Everything stays behind the existing opt-in (`review.metrics: true`, default off). No new gate. `validator-metrics.json` unchanged.

## Design decisions

- **Exact identity key (not fuzzy).** Independent angles may anchor the same issue at slightly different lines / reword past the 40-char stem, so exact keying can undercount. Accepted for determinism + zero false merges; the fuzzy matcher in `intersect-findings.sh` is prosecutor-vs-defender and not reused here.
- **Schema bump reseeds rolling history.** v1→v2 discards the accumulated local aggregate (backed up to `.bak`). Acceptable: per-clone, gitignored, cheaply rebuilt.

## Deferred — per-angle tokens (increment 2)

Token usage is **not capturable on a bare chat host**: the orchestrator model only receives a sub-agent's final message via the `Task` result, not its token usage; the `skills` CLI install never wires a `SubagentStop` hook, and `woostack-init` doesn't touch `.claude/settings.json`. Next cycle will wire CI `action.yml` as the concrete producer plus an optional shipped `capture-usage.sh` + documented hook, degraded-by-default on chat hosts. Full design in the spec §8.

## Tests

- `test-intersect-overlap.sh` (10 assertions): 3-angle cluster math, unanchored exclusion, no self-count, schema v2.
- `test-metrics-fold-overlap.sh` (9 assertions): fold accumulation across runs, v1→v2 reseed.
- Full `woostack-review` + `woostack-init` suites pass; shellcheck clean (only pre-existing SC1091 info).

Spec: `.woostack/specs/2026-06-02-review-metrics-tokens-overlap.md`
Plan: `.woostack/plans/2026-06-02-review-metrics-overlap.md`

### DIFF
--- a/.woostack/config.json
+++ b/.woostack/config.json
@@ -1,0 +1,5 @@
+{
+  "review": {
+    "metrics": true
+  }
+}
--- a/.woostack/plans/2026-06-02-review-metrics-overlap.md
+++ b/.woostack/plans/2026-06-02-review-metrics-overlap.md
@@ -1,0 +1,409 @@
+# Cross-Angle Overlap Metric — Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Add a per-angle cross-angle overlap signal to woostack-review's opt-in metrics so redundant review angles (those whose findings other angles also raise) are visible per run and in the rolling aggregate.
+
+**Architecture:** Extend the existing per-run assembler `emit_angle_metrics` (embedded Python in `intersect-findings.sh`) to cluster raw findings by `(file, line, title_stem)` — excluding unanchored findings — and emit `overlap_total` + `overlap_with` per angle. Extend `metrics-fold.sh` to accumulate those into the rolling `.woostack/metrics.json` and bump its schema version 1→2 (existing version-mismatch path reseeds the old aggregate). Document in `SKILL.md`. Tokens are a separate, deferred increment (spec §8).
+
+**Tech Stack:** Bash + embedded Python 3 + `jq`; bash test harness under `skills/woostack-review/scripts/tests/` sourcing `skills/woostack-init/scripts/tests/assert.sh`.
+
+**Spec:** `.woostack/specs/2026-06-02-review-metrics-tokens-overlap.md` (Increment 1).
+
+---
+
+## File Structure
+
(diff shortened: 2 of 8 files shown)


howarewoo (commented):
## ✅ APPROVED — woostack-review

No findings at or above the configured `severity_floor: high`. Six angles ran (`bugs`, `security`, `conventions`, `aeo`, `tests`, `docs`) through the adversarial prosecutor + defender validator (mode `adversarial`, not degraded).

The swarm raised 5 sub-floor quality nits (1×LOW bugs, 1×MEDIUM conventions, 3×LOW/MEDIUM tests) — all dropped by the `high` floor, so none are posted inline. They're advisory test-coverage and plan-wording suggestions, not blockers.

**Dogfood note (this PR adds the metric being exercised):** `findings.metrics.json` emitted at schema v2 with the new per-angle `overlap_total` / `overlap_with`. All overlap came out `0` — including two angles (`bugs`, `tests`) that flagged the *same* `test-intersect-overlap.sh:44` under *different titles*. That is the exact-key undercount documented in the spec's open questions: location matches but the title-stem differs, so the exact `(file, line, title_stem)` key does not pair them. Working as designed; flagged here as a real-world confirmation.

_Event downgraded APPROVE → COMMENT because the PR is self-authored (GitHub disallows self-approval). Verdict above is the real result._

PR #249, opened 2026-06-07
fix(woostack-review): preflight pending review before posting (#190)

### LINKED ISSUE #190: woostack-review can leave pending draft reviews instead of posting
## Description

`woostack-review` can fail to post its final batched GitHub review and leave a pending draft review behind. On a later run, GitHub then rejects the new review creation with:

```text
Unprocessable Entity: User can only have one pending review per pull request
```

This makes the local review appear complete from the agent side, but the PR does not receive the intended submitted review until someone manually inspects and clears or submits the pending draft.

## Steps to Reproduce

1. Run `/woostack-review` against a PR.
2. Have a posting attempt create or leave a pending review for the authenticated user.
3. Run `/woostack-review` again for the same PR.
4. The create-review API call fails because GitHub only allows one pending review per user per PR.

## Expected Behavior

`woostack-review` should always end in a submitted review or a clearly reported actionable failure. If a pending review already exists, the posting stage should detect it and handle it intentionally, for example by:

- reusing/submitting the existing pending review when safe,
- deleting an empty/stale woostack-owned pending review before retrying,
- refusing with a clear message when the pending review contains unrelated draft comments, or
- using an idempotent marker so it can distinguish woostack-owned draft state from human draft state.

## Actual Behavior

The posting step can fail with HTTP 422 because a pending review already exists. In the observed case, the pending review had an empty summary body but two draft comments (`Fixed in 8bb6a9768.`), so blindly submitting it would have published unrelated draft comments and deleting it would have discarded draft state.

## Observed Context

- Repo reviewed: `howarewoo/TrioSens`
- PR: `howarewoo/TrioSens#1509`
- Existing pending review: `4424013904`
- Failed create-review error: `User can only have one pending review per pull request`
- Manual workaround used: inspect pending review comments, then delete the pending review only after explicit user approval, then repost the prepared payload.
- New posted review after cleanup: `4424079382`

## Suggested Fix

Add pending-review handling to Stage 5 before `POST /repos/{owner}/{repo}/pulls/{pull_number}/reviews`:

1. Query pending reviews for the authenticated user on the PR.
2. If none exist, proceed as today.
3. If a pending review exists and is clearly woostack-owned/stale/empty, delete or submit it according to the desired behavior, then retry once.
4. If it contains non-woostack draft comments or cannot be classified safely, stop with a clear message and do not mutate it automatically.
5. Consider adding a hidden woostack marker to pending review bodies/comments so future runs can distinguish tool-created drafts from human drafts.

## Additional Context

GitHub documents pending reviews as unsubmitted review state and exposes delete/submit endpoints for pending reviews. The tool should use that state model explicitly instead of assuming every create-review call can start from a clean slate.

---
*Filed via Codex using the `file-issue` workflow context.*

howarewoo:
## Goal

Fix #190 — `woostack-review` can leave a pending draft review behind and then 422 on the next run (`User can only have one pending review per pull request`), making the run look complete while the PR receives no submitted review.

## Summary

Adds a pending-review preflight (step 2.5) to the Stage 5 posting procedure in `prompts/_header.md`, before the `POST /repos/{owner}/{repo}/pulls/{pull_number}/reviews` create:

- Queries pending reviews for the authenticated user (`--paginate --slurp`; a `..|objects` jq filter robust to gh-version slurp shape; paginates because the user's own pending review is returned last on busy PRs).
- **None** → proceed as before.
- **Empty + woostack-owned** (hidden `woostack-review:` body marker, 0 draft comments) → DELETE the stale draft, then post fresh (retry once).
- **Anything else** (carries draft comments, or not woostack-owned) → stop with exit 1 and an actionable message naming the draft id + submit/discard commands. No blind submit (would publish unrelated drafts), no blind delete (would discard human work).

A run now always ends in a submitted review or a clearly reported failure — never a silent un-posted state. Documented in the Output Contract (`_header.md`) and Stage 5 (`SKILL.md`).

Single source of truth: all four orchestrator prompts defer to `_header.md` for posting, so the fix lands once and covers the local skill and the CI Action (the Action runs the same prompt; its `Bash(gh:*)` allowlist already permits the new DELETE/GET calls — no allowlist change).

## Test plan

**Automated**

- `bash -n` on the extracted preflight block — clean.
- Ran all three branches against a mocked `gh` with real `jq` (exercises the `..|objects` filter, case-insensitive login match, and `--slurp` nested-array shape):
  - No pending review → proceeds to post.
  - Empty woostack-owned draft (`BotUser` vs `botuser`, marker present, 0 comments) → DELETE called, proceeds.
  - Issue's exact case (empty-body draft, two `Fixed in 8bb6a9768.` comments, id `4424013904`) → stops with exit 1 + actionable message.

**Manual**

- Re-run `/woostack-review` against a PR that has a leftover human pending draft → expect a clear stop message instead of a 422, with commands to resolve the draft.

Fixes #190


howarewoo:
* **#249** <a href="https://app.graphite.com/github/pr/howarewoo/woostack/249?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/howarewoo/woostack/249?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

skills/woostack-review/prompts/_header.md
```diff
@@ -292,6 +294,48 @@ payload = {
...
+# below 422 with "User can only have one pending review per pull request",
+# silently leaving the PR un-reviewed while the run looks complete. Detect any
+# pending review owned by the authenticated user BEFORE posting and resolve it
+# intentionally: discard an empty woostack-owned draft, otherwise stop and ask.
+if [ -n "$AUTH_LOGIN" ]; then
+  AUTH_LC=$(printf '%s' "$AUTH_LOGIN" | tr '[:upper:]' '[:lower:]')
+  # `..|objects` walks every page/array shape `--slurp` may return, so the match
+  # is robust to gh-version differences. A pending review is the user's own and
+  # is returned last, so we MUST paginate to reach it on busy PRs.
+  PENDING=$(gh api "repos/${GITHUB_REPOSITORY}/pulls/$PR_NUMBER/reviews" --paginate --slurp \
```
howarewoo: **Preflight `--slurp` + `--jq` is rejected by current gh**

Step 2.5 runs `gh api ... --paginate --slurp --jq '...'`. On gh 2.92.0 (current, released 2026-04-28) `--slurp` is mutually exclusive with `--jq`: gh exits 1 with "the `--slurp` option is not supported with `--jq` or `--template`" and prints nothing to stdout. The `2>/dev/null || echo ""` swallows that error, so PENDING becomes empty and the `[ -n "$PENDING" ]` guard is never entered. The preflight — the entire purpose of this PR (#190) — silently no-ops on current gh: a leftover pending review is never detected and the step-3 create POST still 422s with "User can only have one pending review per pull request", the exact failure #190 set out to fix. Reproduced by running the exact command against this PR.

Fix: gh forbids `--slurp` together with `--jq`/`--template`. Keep `--slurp` for the all-pages array but move the jq filter to an external process: `PENDING=$(gh api .../reviews --paginate --slurp 2>/dev/null | jq -r "[.. | objects | select((.state? // \"\")==\"PENDING\" and ((.user?.login // \"\") | ascii_downcase)==\"$AUTH_LC\")] | .[0] // empty")`. Add a test that actually executes the preflight against a real gh so this surfaces in CI rather than silently no-opping.

skills/woostack-review/SKILL.md
```diff
@@ -449,6 +449,7 @@ Produces `$OUTDIR/findings.json` — the final validated set — and `$OUTDIR/va
 
 - Build the STATUS_LINE (`APPROVED` / `APPROVED WITH SUGGESTIONS` / `CHANGES REQUESTED`).
 - Submit one `gh api repos/<repo>/pulls/<PR>/reviews` POST containing all inline comments + the summary + status line. The review `event` (`APPROVE` / `COMMENT` / `REQUEST_CHANGES`) is the native gate: any blocking finding (or open prior thread) triggers `REQUEST_CHANGES`; a non-nit non-blocking finding triggers `COMMENT`; nits are event-neutral, so a PR whose only findings are nits gets `APPROVE` with the nits posted inline.
+- Before that POST, the procedure preflights for a leftover **pending review** owned by the authenticated user (GitHub allows only one per user per PR, else the create 422s `User can only have one pending review per pull request`). An empty woostack-owned draft is discarded and the post retried once; any other draft (carrying comments, or not woostack-owned) stops the run with an actionable error instead of being silently mutated. A run thus always ends in a submitted review or a clearly reported failure — never a silent un-posted state.
```
howarewoo: **Nit: Preflight bullet listed after the POST it precedes**

The new bullet at line 452 begins "Before that POST, the procedure preflights..." but it appears after the "Submit one `gh api`..." bullet that describes that POST. A reader scanning the three-bullet list top-to-bottom sees: (1) build STATUS_LINE, (2) submit the POST, (3) before that POST, preflight — which reads as if the preflight follows the submit rather than preceding it. The "Before that POST" qualifier corrects the implied ordering but only if the reader parses the phrase carefully; a quick scan produces the wrong mental model.

Fix: Move the new preflight bullet to appear before the "Submit one `gh api`..." bullet so the listed order matches the execution order: (1) Build STATUS_LINE, (2) Preflight pending review, (3) Submit the POST, (4) DO NOT modify...

<sub>— <strong>LOW · NIT</strong> · flagged by the <code>docs</code> agent</sub>

skills/woostack-review/prompts/_header.md
```diff
@@ -292,6 +294,48 @@ payload = {
...
+    --jq "[.. | objects | select((.state? // \"\")==\"PENDING\" and ((.user?.login // \"\") | ascii_downcase)==\"$AUTH_LC\")] | .[0] // empty" \
+    2>/dev/null || echo "")
+  if [ -n "$PENDING" ]; then
+    PENDING_ID=$(printf '%s' "$PENDING" | jq -r '.id')
+    PENDING_BODY=$(printf '%s' "$PENDING" | jq -r '.body // ""')
+    # Count the draft comments on the pending review. An empty woostack draft is
+    # safe to discard; a draft carrying comments may hold unpublished human work.
+    PENDING_COMMENTS=$(gh api "repos/${GITHUB_REPOSITORY}/pulls/$PR_NUMBER/reviews/$PENDING_ID/comments" \
+      --jq 'length' 2>/dev/null || echo "0")
```
howarewoo: **Nit: Failed comment-count API call defaults to 0, enabling unsafe DELETE**

On line 318, when the `gh api .../reviews/$PENDING_ID/comments` call fails (network error, transient auth issue, etc.), `PENDING_COMMENTS` is set to `"0"` via `|| echo "0"`. The guard on line 320 then treats the pending review as having zero draft comments. If the body also contains the `woostack-review:` marker, the script proceeds to DELETE the pending draft — even though it may actually contain draft comments that the API call failed to retrieve. This inverts the stated safety invariant: a failed count-check silently chooses the more destructive path (delete) rather than the conservative path (stop and report). A draft that holds unpublished human work could be silently discarded whenever the comments endpoint is transiently unreachable.

Fix: Change the fallback from `|| echo "0"` to `|| echo "ERR"` (or any non-numeric sentinel). Then update the guard condition so it is true only when the count was definitely 0, and add a separate branch for the sentinel value that exits 1 with a message like "could not determine pending review comment count — aborting to avoid data loss". This ensures the script fails safe (error out) when the comment count cannot be determined, rather than assuming an empty draft.

<sub>— <strong>MEDIUM · NIT</strong> · flagged by the <code>bugs</code> agent</sub>

howarewoo (commented):

Reviewed the pending-review preflight added for #190 across 6 angles (bugs, security, conventions, aeo, docs, skills) with adversarial prosecutor+defender validation. **Blocking:** the preflight's `gh api --paginate --slurp --jq` command is rejected by current gh (2.92.0 — `--slurp` is mutually exclusive with `--jq`), so with `2>/dev/null` the preflight silently no-ops and the fix does not take effect on a current gh. Two non-blocking nits noted inline.

---
**Status: CHANGES REQUESTED** — 1 blocking + 0 non-blocking finding(s) (1 HIGH, 0 MEDIUM, 0 LOW) + 2 nit(s). See inline comments.



skills/woostack-review/prompts/_header.md
howarewoo: Fixed in 40f15d9. `--slurp` and `--jq`/`--template` are mutually exclusive in gh (it exits non-zero), so the combined call under `2>/dev/null` silently no-opped the whole preflight. Moved the filter to an external `jq` on the slurped array and documented the constraint inline. Verified the corrected command runs (exit 0) against this PR.

skills/woostack-review/SKILL.md
howarewoo: Fixed in 40f15d9. Reordered the Stage 5 bullets to STATUS_LINE -> preflight -> submit POST -> DO-NOT so the listed order matches execution order.

skills/woostack-review/prompts/_header.md
howarewoo: Fixed in 40f15d9. A failed comment-count now yields a non-numeric sentinel (`ERR`) and aborts (exit 1) rather than defaulting to `0`; the DELETE path is reached only when the count is definitely known, so a transient error can no longer discard an unread draft.

skills/woostack-review/prompts/_header.md
```diff
@@ -292,6 +294,60 @@ payload = {
...
+      --jq 'length' 2>/dev/null || echo "ERR")
+    if [ "$PENDING_COMMENTS" = "ERR" ]; then
+      echo "ERROR: could not determine the draft-comment count for pending review $PENDING_ID on ${GITHUB_REPOSITORY}#$PR_NUMBER — aborting to avoid discarding unpublished work. Resolve the draft manually and re-run." >&2
+      exit 1
+    fi
+    # woostack-owned <=> our hidden body marker is present (see review body, step 1).
+    if printf '%s' "$PENDING_BODY" | grep -q 'woostack-review:' && [ "$PENDING_COMMENTS" = "0" ]; then
+      # Empty, woostack-owned stale draft: discard it, then post fresh (retry once).
+      echo "woostack-review: discarding empty stale pending review $PENDING_ID before posting." >&2
+      gh api --method DELETE "repos/${GITHUB_REPOSITORY}/pulls/$PR_NUMBER/reviews/$PENDING_ID"
```
howarewoo: **Nit: Unguarded DELETE lets failed discard fall through to POST**

Line 335 issues the stale-draft DELETE with no error check. If the DELETE fails (network error, 403, or the review was submitted between detection and delete), gh exits non-zero but — because set -e is not in effect and there is no || exit 1 — execution falls through to step 3's POST. That POST then 422s with "User can only have one pending review per pull request", the same opaque failure the preflight was introduced to prevent. The caller sees a cryptic 422 instead of the actionable message that was deliberately written for the stop path.

Fix: Add || { echo "ERROR: failed to delete pending review $PENDING_ID — aborting." >&2; exit 1; } after the DELETE call, consistent with the explicit exit 1 guards already on lines 329 and 346.

```suggestion
      gh api --method DELETE "repos/${GITHUB_REPOSITORY}/pulls/$PR_NUMBER/reviews/$PENDING_ID" \
        || { echo "ERROR: failed to delete pending review $PENDING_ID on ${GITHUB_REPOSITORY}#$PR_NUMBER — resolve the draft manually and re-run." >&2; exit 1; }
```

<sub>— <strong>MEDIUM · NIT</strong> · flagged by the <code>bugs</code> agent</sub>

howarewoo (commented):

Re-review of head 40f15d9 after the fixes. The prior **blocking** finding (the `gh --slurp`/`--jq` preflight no-op) is **resolved** — the corrected external-`jq` form was executed and runs clean (exit 0) on current gh; the two prior nits are also fixed. One new non-blocking nit: the discard `DELETE` has no error guard, so a failed delete falls through to the create POST and 422s with the same opaque error. Net: approved.

---
**Status: APPROVED** — No blocking findings, 1 nit(s). See inline comments.



skills/woostack-review/prompts/_header.md
howarewoo: Fixed in 80da786. Added `|| { echo "ERROR: failed to delete pending review ..." >&2; exit 1; }` after the discard DELETE, so a failed delete aborts with the actionable message instead of falling through to the step-3 POST and 422ing. Consistent with the other exit-1 guards in the block. Verified the full preflight block with `bash -n` (clean).

Issue #321, opened 2026-06-12
Make woostack-review OUTDIR per-run by default


howarewoo:
## Problem

Local `woostack-review` runs default to a per-project `OUTDIR` (`.woostack/tmp/pr-review-<repo-hash>` or `/tmp/pr-review-<repo-hash>`). That makes repeated or concurrent reviews of the same repo share the same artifact tree.

During a local `/woostack-review 317` run, prefetch selected the per-project directory and warned that it held in-flight `findings.*` artifacts. The run then continued with that directory still active. Receipt verification saw stale receipts/findings from a prior review, including unrelated `site/` findings, even though PR 317 only enabled six angles.

I caught this before posting, discarded the contaminated run, and restarted with an explicit per-run `OUTDIR`:

```text
.woostack/tmp/pr-review-317-20260612172146-54329
```

The isolated rerun produced the correct review.

## Why This Matters

The current default makes the safest local workflow depend on the orchestrator remembering to set a unique `OUTDIR`. If it forgets, stale artifacts can leak into merge/validation and produce false review comments. The receipt gate helps expose the problem when extra receipts exist, but the system should avoid the contaminated directory in the first place.

## Proposed Fix

Make local chat-host review runs use a per-run `OUTDIR` by default, while preserving explicit `OUTDIR` overrides.

Possible shape:

- `resolve-outdir.sh` keeps the stable per-project helper available for scripts that need it.
- `prefetch.sh` or the local orchestrator path derives a per-run directory when `GITHUB_ACTIONS != true`, e.g. `.woostack/tmp/pr-review-<pr-or-local>-<timestamp>-<pid>`.
- If `prefetch.sh` refuses to wipe a directory because it detects in-flight findings, it should hard-stop instead of leaving that same contaminated `OUTDIR` active.
- Stage docs should tell local hosts to export the resolved per-run `OUTDIR` to every sub-agent.

## Acceptance Criteria

- Two local reviews of the same repo do not share findings/receipt artifacts unless the caller explicitly sets the same `OUTDIR`.
- A stale or in-flight directory cannot proceed to merge/validation/posting after a refused wipe.
- Existing explicit `OUTDIR` behavior remains supported for sandboxed hosts and tests.
- GitHub Action behavior remains deterministic and does not leak artifacts across jobs.

## Observed Incident

- Command: `/woostack-review 317`
- Expected angles: `bugs`, `security`, `conventions`, `aeo`, `tests`, `docs`
[...]

PR #317, opened 2026-06-12
fix(review): resolve self-path under zsh-sourced scripts (#314)

howarewoo:
## Goal

Fix `resolve-outdir.sh` mis-resolving `OUTDIR` to `sha1("")` (`/tmp/pr-review-da39a3ee5e6b`) when a host **sources** woostack-review/address-comments scripts from a non-bash shell (zsh — Claude Code's default Bash-tool shell on macOS), so review artifacts land in the real per-project tree instead of a shared empty-string dir that silently reuses stale findings (#314).

## Summary

- Hardened the `${BASH_SOURCE[0]}` self-path **bug class** to `${BASH_SOURCE[0]:-$0}` (the idiom already shipped at `woostack-review/SKILL.md:223`) across every path-resolution site in `woostack-review/scripts/**` and `woostack-address-comments/scripts/**` — 18 scripts, incl. both `resolve-outdir.sh` copies (the one site that actually manifested).
- Left `resolve-model.sh:126` `[ "${BASH_SOURCE[0]}" = "${0}" ]` bare on purpose — it is the dual-mode execution guard; `:-$0` would make it always-true under zsh and run `main` on every source.
- Added `skills/woostack-review/scripts/tests/test-resolve-outdir-zsh.sh`: sources both copies under real `zsh` and asserts `WOOSTACK_ROOT`/`OUTDIR` derive from the real root (not the empty-string hash); also statically pins "no production script uses a bare `dirname "${BASH_SOURCE[0]}"`". Skips gracefully when `zsh` is absent.
- Distilled the gotcha to `.woostack/memory/` and corrected the sibling note that recommended the bare idiom.

### Automated

- [x] `bash skills/woostack-review/scripts/tests/test-resolve-outdir-zsh.sh` → 10 passed, 0 failed (was 0/10 before the fix — red→green).
- [x] All review + address-comments suites: 36 passed, 0 failed.
- [x] `bash -n` on all 18 modified scripts + the new test → clean.
- [x] `shellcheck -S warning` on `resolve-outdir.sh` + new test → clean.

### Manual

**Before merge**

- [ ] From a non-`scripts/` cwd: `zsh -c 'unset OUTDIR WOOSTACK_ROOT; source <repo>/skills/woostack-review/scripts/resolve-outdir.sh; echo $OUTDIR $WOOSTACK_ROOT'` → `OUTDIR` is a real-root hash (not `da39a3ee5e6b`), `WOOSTACK_ROOT` non-empty, no `source:17` error.
[...]

Issue #314, opened 2026-06-12
resolve-outdir.sh resolves OUTDIR to sha1('') under a zsh host (empty ${BASH_SOURCE[0]})


howarewoo:
## Summary

`scripts/resolve-outdir.sh` mis-resolves `OUTDIR` to a hash of the **empty string** when the host's shell tool is **zsh** (or any non-bash), because `${BASH_SOURCE[0]}` is empty there. Every downstream stage then reads/writes the wrong `OUTDIR` (`/tmp/pr-review-da39a3ee5e6b`, where `da39a3ee5e6b` = `sha1("")[:12]`), and a stale tree at that path is silently reused.

This bites any host that drives the skill through a zsh shell (e.g. Claude Code's Bash tool on macOS defaults to zsh), following the documented Stage 1 step `source resolve-outdir.sh`.

## Root cause

`scripts/resolve-outdir.sh:17`:

```bash
source "$(dirname "${BASH_SOURCE[0]}")/resolve-root.sh"
```

- Under zsh, `${BASH_SOURCE[0]}` is unset/empty → `dirname ""` → `.` → it tries `./resolve-root.sh`, which only exists if CWD happens to be the scripts dir.
- When CWD is elsewhere (e.g. a git worktree), the `source` fails:
  ```
  resolve-outdir.sh:source:17: no such file or directory: ./resolve-root.sh
  ```
  `resolve-root.sh` never runs → `WOOSTACK_ROOT` stays empty → `printf '%s' "" | sha1sum` → `da39a3ee5e6b…` → `OUTDIR=/tmp/pr-review-da39a3ee5e6b`.
- Worse: if a host exports that broken `OUTDIR`, subsequent `bash prefetch.sh` calls **inherit** it, so even the bash-shebang'd scripts land in the wrong tree.

## Repro

On a zsh host, from any directory that is not the skill's `scripts/` dir:

```zsh
source /path/to/woostack-review/scripts/resolve-outdir.sh
# -> resolve-outdir.sh:source:17: no such file or directory: ./resolve-root.sh
echo "$OUTDIR"
# -> /tmp/pr-review-da39a3ee5e6b   (sha1 of empty string)
```

## Impact

- Review artifacts land in a per-empty-string `/tmp` dir instead of `${WOOSTACK_ROOT}/.woostack/tmp/pr-review-<realhash>`.
[...]

PR #323, opened 2026-06-12
fix(review): per-run OUTDIR default + local hard-stop on contaminated dir (#321)

howarewoo:
## Goal

Local `woostack-review` runs of the same repo were sharing one per-project `OUTDIR`, so stale `findings.*`/`receipt.*` from a prior run could leak into merge/validation/posting and produce false review comments (issue #321). Make the local default per-run and stop a contaminated directory from proceeding.

## Summary

- `resolve-outdir.sh`: local default is now a per-**run** `pr-review-<hash>-<ts>-<pid>` dir (`<hash>` isolates repos, `<ts>-<pid>` isolates runs). Under `GITHUB_ACTIONS=true` it keeps the stable per-project `pr-review-<hash>` form; explicit `OUTDIR` overrides are still honored verbatim.
- `prefetch.sh`: the in-flight-`findings.*` guard now **hard-stops locally** (`::error::` + `exit 1`) instead of warn-and-continue. In CI it still **preserves** the validate job's legitimately-downloaded `findings.*` and continues (`GITHUB_ACTIONS`-gated). `WOO_REVIEW_FRESH=1` forces a wipe either way.
- `SKILL.md` Stage 1: documents the per-run default, the capture-`outdir=`-and-export contract, and the local hard-stop / CI-preserve split.
- Tests: new `test-resolve-outdir-per-run.sh` (per-run isolation, explicit override, CI determinism, 3-way prefetch guard); `test-resolve-outdir-zsh.sh` updated for the per-copy shape divergence (review per-run, address-comments unchanged).

`woostack-address-comments`'s copy of `resolve-outdir.sh` is intentionally left per-project — it has no wipe/merge step and thus no contamination hazard.

### Automated

- [x] `bash skills/woostack-review/scripts/tests/test-resolve-outdir-per-run.sh` → 13 passed, 0 failed
- [x] `bash skills/woostack-review/scripts/tests/test-resolve-outdir-zsh.sh` → 10 passed, 0 failed
- [x] Full review test suite (`for t in test-*.sh`) → 33 passed, 0 failed
- [x] `shellcheck resolve-outdir.sh prefetch.sh` → no new findings vs. baseline (pre-existing SC1091/SC2126 info only)

### Manual

**Before merge**

- [ ] Read the diff: confirm only `woostack-review` paths change; `woostack-address-comments` untouched.
- [ ] Run two local `/woostack-review` passes on the same repo without setting `OUTDIR`; confirm each lands on a distinct `pr-review-<hash>-<ts>-<pid>` dir.
- [ ] Seed a stale `findings.bugs.json` into a reused `OUTDIR` and run `prefetch.sh` locally; confirm it aborts with `::error::` and does not wipe.

**After merge**

- [ ] Confirm a real GitHub Action review run is unaffected: detect/review/validate jobs complete and the validator reads the downloaded `findings.*` (CI-preserve path).

Note: stacked on #317 (issue #314) — both edit `resolve-outdir.sh` and `test-resolve-outdir-zsh.sh`; basing on `main` would conflict.

Spec: .woostack/fixes/2026-06-12-review-outdir-per-run.md


howarewoo:
* **#323** <a href="https://app.graphite.com/github/pr/howarewoo/woostack/323?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/howarewoo/woostack/323?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

skills/woostack-review/scripts/tests/test-resolve-outdir-per-run.sh
```diff
@@ -0,0 +1,129 @@
...
+# WOO_REVIEW_TEST_MODE is refused under GITHUB_ACTIONS, so the no-PR early skip
+# (not the fake-data hooks) is what lets this exit cleanly.
+env -u WOO_REVIEW_FRESH -u PR_NUMBER GITHUB_ACTIONS=true GITHUB_REPOSITORY=owner/repo \
+  OUTDIR="$seed_b" bash "$PREFETCH" >"$out_b_run" 2>"$err_b"
+rc_b=$?
+set -e
+assert_exit 0 "$rc_b" "CI prefetch does not abort at the guard (continues past it)"
+assert_exit 0 "$([ -e "$seed_b/findings.bugs.json" ]; echo $?)" \
+  "CI preserve keeps the downloaded findings.* in place"
+assert_not_contains "$(cat "$err_b")" "contaminated" \
```
howarewoo, line 88: **Nit: Test 4b missing positive assertion for CI warning**

Test 4b (CI preserve path) only asserts that `"contaminated"` is absent from stderr — it does not assert that the expected `::warning::` preservation message was actually emitted. A regression that silently swallows the CI warning (or emits nothing to stderr) would still pass all three assertions in test 4b (exit 0, findings.bugs.json still present, no "contaminated" in stderr). The intent expressed in the assertion message — "CI path uses a warning, not the local hard-stop error" — is only half-pinned: the "not local hard-stop" half is checked, but the "CI warning WAS emitted" half is not.

Fix: Add a positive assertion after line 88 that the CI warning message was emitted: `assert_contains "$(cat "$err_b")" "preserving" "CI guard emits the preserve warning message"`. The word `"preserving"` is a stable ASCII substring of the production warning string ("preserving (CI validate path)") and directly pins the distinction from the local hard-stop path.

```suggestion
assert_not_contains "$(cat "$err_b")" "contaminated" \
  "CI path uses a warning, not the local hard-stop error"
assert_contains "$(cat "$err_b")" "preserving" \
  "CI guard emits the preserve warning message"
```

skills/woostack-review/SKILL.md
```diff
@@ -228,27 +228,29 @@ export WOO_REVIEW_ACTION_PATH="$(cd "$(dirname "${BASH_SOURCE[0]:-$0}")" && pwd)
...
 ```bash
-# Resolve the per-project OUTDIR once and export it so prefetch.sh and every
-# sub-agent share one tree. Default: /tmp/pr-review-<hash> derived from the git
-# toplevel (scripts/resolve-outdir.sh), so different repos on one machine never
-# collide. An explicit OUTDIR (e.g. a sandbox temp dir) is respected as-is.
+# Resolve OUTDIR once and export it so prefetch.sh and every sub-agent share one
+# tree. Local default: a per-RUN pr-review-<hash>-<ts>-<pid> dir
+# (scripts/resolve-outdir.sh), so two reviews of the same repo never collide. An
+# explicit OUTDIR (e.g. a sandbox temp dir) is respected as-is. Because the
+# per-run value is non-deterministic, capture prefetch's printed outdir=<path>
```
howarewoo: **Nit: Code comment says capture outdir but example doesn't show it**

The comment on lines 244-245 tells hosts to "capture prefetch's printed outdir=<path> and re-export OUTDIR from it before fanning out sub-agents (no recompute drift)" but the code example that follows only shows `bash prefetch.sh` with no capture step. A host agent reading the code block and following it literally will call prefetch without capturing its stdout output. The prose callout at line 233 states this is a MUST, so the code example and the prose are inconsistent: the comment describes a step the example does not demonstrate.

Fix: Either add a concrete capture line to the code example (e.g. `OUTDIR="$(bash ... prefetch.sh | tee /dev/stderr | awk -F= '/^outdir=/{print $2}')"` pattern, or a `read -r` idiom), or rephrase the comment to clarify that the existing export from `source resolve-outdir.sh` on the prior line already sets OUTDIR and the capture is only needed when those two steps are separated. The current comment overstates the requirement relative to what the example shows.

<sub>— <strong>LOW · NIT</strong> · flagged by the <code>skills</code> agent</sub>

howarewoo (commented):

This PR makes the local default OUTDIR per-run instead of per-project and adds a hard-stop that refuses to reuse a contaminated artifact directory, with new shell tests covering the local hard-stop, the CI preserve path, and the `WOO_REVIEW_FRESH=1` force-wipe. The core logic, security (no injection/path-traversal/unsafe `rm -rf` surface), and conventions all check out — only two LOW nits on doc/example consistency and a test-assertion gap.

---
**Status: APPROVED** — No blocking findings, 2 nit(s). See inline comments.



skills/woostack-review/scripts/tests/test-resolve-outdir-per-run.sh
howarewoo: Fixed in `91e05be`. Added `assert_contains "$(cat "$err_b")" "preserving"` to test 4b — the CI preserve path now positively asserts the `::warning::...preserving (CI validate path)` message was emitted, not just that the local hard-stop's `contaminated` error was absent. Test suite green (14 passed).

skills/woostack-review/SKILL.md
howarewoo: Fixed in `91e05be`. Reworded the code-block comment instead of editing the example: the `source resolve-outdir.sh` line already sets+exports OUTDIR in that shell, so prefetch and same-shell sub-agents inherit it verbatim. The `outdir=<path>` capture is now framed as the fallback only for when prefetch runs in a non-inherited subprocess — matching what the example actually shows.

PR #325, opened 2026-06-13
fix(execute): self-pin subagent implementer to its worktree (cwd-less hosts)

### LINKED ISSUE #324: Subagent driver can't honor the worktree contract on hosts whose spawn API lacks a cwd parameter
## Summary

The **subagent driver** for `woostack-execute` / `woostack-execute-overnight` cannot honor the worktree contract on hosts whose subagent-spawn API has **no `cwd` parameter**. The contract requires the controller to dispatch each implementer with `cwd = $wt` (the per-PR worktree). When the host can't set a subagent's cwd, subagent-mode implementation is effectively blocked — the only safe path is to fall back to the inline driver.

## The contract requirement

`woostack-init/references/worktrees.md` §2 (Operate):

> From the first write onward the run operates with **cwd = `$wt`** … any sub-skill the run calls … inherits that cwd so it authors into the worktree, not the primary tree. **In subagent mode the controller dispatches implementers with cwd = `$wt`.**

§3 (Hard invariant):

> A run does all its writes in its own worktree; the primary checkout stays on the base branch, clean … **the primary tree is never edited.**

So subagent mode is *defined* in terms of a per-implementer cwd the controller sets.

## Why it's blocked on a cwd-less spawn API

Concrete host: Claude Code's `Agent` tool. Its spawn schema has `subagent_type`, `model`, `prompt`, `isolation`, `run_in_background` — **no `cwd`**. The two reachable behaviors both violate the contract:

1. **Inherit parent cwd** → the subagent runs in the **primary checkout** (on the protected base branch, e.g. `staging`). Any write there violates §3 ("the primary tree is never edited") and dirties the tree all other runs branch from.
2. **`isolation: "worktree"`** → the host creates a *fresh throwaway* worktree, **not** the controller's tracked per-PR branch worktree (`$wt`, created + `gt track --parent <base>`). Commits land in a detached/unintended worktree, not on the PR branch.

Neither equals the contract's required `cwd = $wt`. So the smart default ("subagent where the host can spawn subagents, else inline") resolves to **subagent** — the host *can* spawn agents — but the driver then can't faithfully place those agents. The honest move is to override to inline, which is what an unattended run must do to avoid a subagent editing the protected primary tree.

## Impact

- On any cwd-less-spawn host, `--subagent` (and the subagent smart-default) is unsafe for implementation: a dispatched implementer can silently edit the primary `staging`/`main` checkout.
- Forces inline driver, losing the subagent driver's isolation/parallelism for the build phase. (Review-swarm subagents are unaffected — they're read-only and write only to `$OUTDIR`.)
- The failure is silent unless the controller knows to override; nothing in the dispatch path asserts "I'm actually in `$wt`."

## Suggested fixes

1. **Dispatch-prompt cwd pinning + guard (works without a `cwd` param).** Have the subagent-driver embed, as the implementer's first actions: `cd "$wt"` then a hard assertion `[ "$(git rev-parse --show-toplevel)" = "$wt" ] || exit 1` **before any write**. This lets a cwd-less host run subagent mode safely — the implementer self-pins and aborts if it isn't in the worktree, so it can never write to the primary tree.
2. **Capability check + documented fallback.** `subagent-driver.md` should detect whether the host's spawn primitive supports a per-call cwd; if not, state plainly that inline is the required fallback (today this is left for the controller to infer).
3. **Host guidance.** Recommend hosts expose `cwd` on their agent-spawn primitive; note `isolation: "worktree"` is *not* a substitute (it makes a new worktree, not the tracked PR branch).

Fix #1 is the most portable — it removes the blocker without depending on the host gaining a `cwd` param.

## Workaround today

Use the inline driver (`--inline`, or let the smart-default fall back) on hosts without a per-subagent cwd. Reported honestly in the run's decision log.

---
Filed by Claude Code after an unattended `woostack-execute-overnight` run: smart-default resolved to subagent, but the `Agent` tool has no `cwd` parameter, so the run fell back to inline to keep implementers out of the protected primary `staging` tree.

howarewoo:
## Goal

Unblock subagent-driven `woostack-execute` / `woostack-execute-overnight` on hosts whose spawn API has no per-call `cwd` (e.g. Claude Code's `Agent` tool). Today the worktree contract defines subagent isolation as a controller-set `cwd = $wt`, which such hosts can't honor — a dispatched implementer silently inherits the primary checkout and can write to the protected base tree. Closes #324.

## Summary

- `prompts/implementer.md`: add a "Worktree pin (do this FIRST)" block before `## Task` — the implementer `cd`s into `$wt` and hard-asserts `git rev-parse --show-toplevel == $wt` (path-normalized via `pwd -P`), aborting **before any write** if it isn't in the worktree.
- `references/subagent-driver.md`: dispatch step 1 fills the prompt pin with `$wt` and sets a per-call cwd where the host exposes one; new **Worktree placement** section with a capability table (per-call-cwd / cwd-less self-pin / no-shell→inline) and states `isolation:"worktree"` is **not** a substitute (throwaway worktree, not the tracked PR branch).
- `worktrees.md` §2 and `woostack-execute/SKILL.md` step 1: amend the `cwd = $wt` lines to describe the two-layer (per-call-cwd + always-on self-pin) guard.
- Distilled gotcha note `subagent-self-pins-to-worktree.md` (+ index).
- Overnight reuses the same driver + contract files, so the fix covers its `--subagent` and smart-default paths centrally.

### Automated

- [x] Structural checks (no-runner doc edit): guard present in `implementer.md`, pin block before `## Task`, `## Worktree placement` + "not a substitute" in driver, `show-toplevel` in `worktrees.md`, "self-pinned by the dispatch-prompt guard" in `SKILL.md`, no stale "dispatches implementers with cwd" — 6/6 PASS.
- [x] `woostack-init/doctor.sh` — 0 errors (1 pre-existing overlap-cluster warning, unrelated).
- [x] Cross-file relative links + `#worktree-placement` anchor resolve.

### Manual

**Before merge**

- [ ] Read the diff: confirm the pin block precedes any write step and the assert aborts (no fallthrough) on mismatch.
- [ ] On a cwd-less-spawn host, run `/woostack-execute <plan> --subagent` and confirm implementer writes land only in `$wt`, and that a deliberately wrong `$wt` makes the implementer abort BLOCKED before writing.

Spec: .woostack/fixes/2026-06-12-subagent-cwd-pin.md


howarewoo:
* **#325** <a href="https://app.graphite.com/github/pr/howarewoo/woostack/325?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/howarewoo/woostack/325?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

skills/woostack-execute/references/subagent-driver.md
```diff
@@ -25,13 +25,43 @@ There is **no per-task git commit.** Each implementer leaves its work uncommitte
...
+
 ## Per-task loop
 
 For each task in the increment, in order:
 
 1. **Dispatch an implementer subagent** with [../prompts/implementer.md](../prompts/implementer.md).
    Pass the full task text and exactly the context it needs — the subagent never inherits this
-   session's history. Resolve and pass its model per
+   session's history. **Place it in the increment's worktree per [Worktree placement](#worktree-placement)
+   below:** fill the prompt's `<worktree absolute path — $wt>` pin with `$wt`, and additionally set
```
howarewoo: **Nit: "below" should be "above" in Worktree placement cross-ref**

Per-task loop step 1 (line 63) reads "per [Worktree placement](#worktree-placement) below:" but the ## Worktree placement section is at line 28, which is above the ## Per-task loop section (line 56). The anchor link resolves correctly, but the directional word "below" misdirects a reader who will scroll down and not find the section there.

Fix: Replace "below:" with "above:" so the directional hint matches the actual document layout.

```suggestion
   above:** fill the prompt's `<worktree absolute path — $wt>` pin with `$wt`, and additionally set
```

<sub>— <strong>LOW · NIT</strong> · flagged by the <code>docs</code> agent</sub>

howarewoo (commented):

Reviewed the cwd-less-host self-pin fix across woostack-execute and woostack-init docs. The change is internally consistent and well cross-linked; the only finding is one directional-reference nit. Approved.

---
**Status: APPROVED** — No blocking findings, 1 nit. See inline comments.



skills/woostack-execute/references/subagent-driver.md
howarewoo: Fixed in dfcbf44 — changed `below:` to `above:` so the directional hint matches the layout (Worktree placement is above Per-task loop).

PR #352, opened 2026-06-14
fix(status): stop flagging the spec+plan PR at `ready` as drift

howarewoo:
## Goal

Stop `/woostack-status` from raising a false **"status lags"** drift when the spec+plan
handoff PR opens while the plan is still at `status: ready` — the documented, expected state
between plan-harden and execution.

## Summary

- `ready` sat in the lag-flag head-state list (`status.sh`), so any discovered PR at `ready`
  (`prcount ≥ 1`) tripped a "status lags" drift. But `conventions.md` and the `ready`
  `next_action` both say the spec+plan PR is opened **at** `ready`, before execution — the
  flag contradicted the convention that blesses it. Introduced together with the `ready`
  phase in `2026-06-12-build-loop-ready-phase.md`.
- **Fix:** drop `ready` from the lag `case`; flag only the genuinely pre-PR head states
  (`draft`/`hardened`/`approved`/`planning`).
- **Latent test bug:** the guard named *"ready with open PR flagged"* never tested `ready` —
  `mkplan` defaulted the plan to `planning`, so the real `ready`+PR path shipped untested.
  Replaced with a real `ready` test (plan `status: ready`) asserting **no** drift, plus an
  explicit `planning`+PR guard so the genuine-lag nudge stays covered.
- Reworded the `conventions.md` drift bullet to name `ready` as exempt.
- **Non-goal:** `resolve_phase` still derives `eff=in-review` when any PR is OPEN — pre-existing
  behavior, not the reported symptom (the drift flag). Left unchanged.

## Test plan

**Automated**
- `bash skills/woostack-status/scripts/tests/run-tests.sh` → **66 passed, 0 failed**.
- Red→Green confirmed: the new `ready`+spec+plan-PR assertion fails before the `status.sh`
  edit (flag fires), passes after.
- `planning`+PR and `approved`+PR lag tests still pass → genuine-lag coverage intact.

**Manual**
- `grep -n ready skills/woostack-status/scripts/status.sh` → `ready` retained in
  `VALID_PHASES` and `next_action`, absent only from the lag `case`.

Spec: .woostack/fixes/2026-06-14-status-ready-pr-drift.md


howarewoo:
* **#352** <a href="https://app.graphite.com/github/pr/howarewoo/woostack/352?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/howarewoo/woostack/352?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

howarewoo (commented):
woostack-review: APPROVED

No validated findings.

Note: authenticated reviewer is the PR author, so GitHub cannot record this as a native APPROVE event; posting as COMMENT.

PR #385, opened 2026-06-16
fix(woostack-status): exclude fenced example checkboxes from plan progress

howarewoo:
## Affected spec sections

- [ ] `spec/architecture.md`
- [ ] `spec/frameworks.md`
- [ ] `spec/infrastructure.md`
- [ ] `spec/patterns.md`
- [ ] `spec/development.md`
- [ ] `spec/bootstrap.md`
- [ ] `README.md` / other top-level docs

## Checklist

- [ ] Cross-links to related sections still resolve
- [ ] No version numbers hard-coded in `frameworks.md` where "latest" suffices

### DIFF
--- a/skills/woostack-status/scripts/tests/test-status.sh
+++ b/skills/woostack-status/scripts/tests/test-status.sh
@@ -71,6 +71,27 @@ run_status "$p"
 assert_contains "$OUT" "3/10" "plan progress counted"
 assert_contains "$OUT" "harden the plan" "planning next-action"
 
+# Fenced example checkboxes (embedded template / SKILL.md literals) must NOT count
+# as tasks. Real: 3 done + 3 todo = 6. The 5 boxes inside the ```` fence (including
+# a nested ``` block, which must not close the enclosing fence) are ignored.
+fence="$(mktemp -d)/.woostack"
+mkspec "$fence" echo planning feature/echo
+mkdir -p "$fence/plans"
+{
+  printf -- '---\ntype: plan\nsource: .woostack/specs/2026-06-01-echo.md\nstatus: planning\nbranch: feature/echo\n---\n\n'
+  printf -- '**Source:** .woostack/specs/2026-06-01-echo.md\n\n'
+  printf -- '- [x] real done 1\n- [x] real done 2\n- [x] real done 3\n- [ ] real todo 1\n- [ ] real todo 2\n\n'
+  printf -- '````markdown\n'
+  printf -- '- [ ] embedded example a\n- [ ] embedded example b\n'
+  printf -- '```bash\n- [ ] nested fenced c\n```\n'
+  printf -- '- [ ] embedded example d\n- [ ] embedded example e\n'
+  printf -- '````\n\n'
+  printf -- '- [ ] real todo 3 after fence\n'
+} > "$fence/plans/2026-06-01-echo.md"
+run_status "$fence"
+assert_contains "$OUT" "3/6" "fenced example checkboxes excluded from progress"
+assert_not_contains "$OUT" "3/11" "fenced boxes not over-counted"
+
 # Source line as an Obsidian wikilink ([[specs/<basename>]]) resolves the plan, same as a
 # bare path. The plan basename intentionally differs from the spec slug so resolution can
 # only come from the **Source:** line, not the slug fallback.
--- a/skills/woostack-status/scripts/tests/test-status.sh
+++ b/skills/woostack-status/scripts/tests/test-status.sh
@@ -92,6 +92,48 @@ run_status "$fence"
 assert_contains "$OUT" "3/6" "fenced example checkboxes excluded from progress"
 assert_not_contains "$OUT" "3/11" "fenced boxes not over-counted"
 
+# A plain triple-backtick fence (the everyday variant) must also exclude its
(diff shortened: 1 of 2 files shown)


howarewoo:
* **#385** <a href="https://app.graphite.com/github/pr/howarewoo/woostack/385?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/howarewoo/woostack/385?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

skills/woostack-status/scripts/tests/test-status.sh
```diff
@@ -71,6 +71,27 @@ run_status "$p"
...
+  printf -- '- [x] real done 1\n- [x] real done 2\n- [x] real done 3\n- [ ] real todo 1\n- [ ] real todo 2\n\n'
+  printf -- '````markdown\n'
+  printf -- '- [ ] embedded example a\n- [ ] embedded example b\n'
+  printf -- '```bash\n- [ ] nested fenced c\n```\n'
+  printf -- '- [ ] embedded example d\n- [ ] embedded example e\n'
+  printf -- '````\n\n'
+  printf -- '- [ ] real todo 3 after fence\n'
+} > "$fence/plans/2026-06-01-echo.md"
+run_status "$fence"
+assert_contains "$OUT" "3/6" "fenced example checkboxes excluded from progress"
```
howarewoo, line 92: **Nit: Simple 3-backtick fence has no standalone test**

The new fence-exclusion test only exercises a 4-backtick opening fence (````markdown) with a nested 3-backtick block inside whose job is to prove it does NOT close the outer fence. The most common fence variant — a plain triple-backtick fence that opens and closes independently — is never tested as the primary fence under scrutiny. The awk logic that handles it (`flen=3`, closing with `run>=flen`) shares the same code path and is correct, but regression coverage is absent for the everyday case. A future refactor that changes the `run>=flen` comparison (e.g. to `run==flen`) would break 3-tick fences while this test still passes.

Fix: Add a second fixture that uses a plain triple-backtick fence (``` rather than ````) around a block of example checkboxes, assert the real tasks outside the fence count correctly (e.g. assert_contains "$OUT" "2/3" for 1 done + 1 todo outside and 2 fenced), and assert the fenced boxes are not counted.

skills/woostack-status/scripts/tests/test-status.sh
```diff
@@ -71,6 +71,27 @@ run_status "$p"
...
+assert_not_contains "$OUT" "3/11" "fenced boxes not over-counted"
```
howarewoo, line 93: **Nit: Unclosed fence edge case not tested**

If a plan file contains an opening fence (e.g. ```markdown) with no matching closing fence — a plausible state for an agent-authored or mid-write plan — the awk's `infence` variable stays 1 for the rest of the file and all remaining real task checkboxes are silently excluded. This under-counts `t` (total tasks) and makes the plan appear more complete than it is (e.g. a feature with 3 done and 2 todo after an unclosed fence reports `3/3` instead of `3/5`). No test covers the unclosed-fence path, so the silent under-count behavior is unguarded against future changes to the awk logic.

Fix: Add a fixture with an unclosed fence followed by real task checkboxes, and decide whether to (a) count remaining tasks after an unclosed fence (treating them as real tasks — document this choice), or (b) preserve current behavior (exclude them) but assert it explicitly so any change is visible. The test should assert a specific count, e.g. assert_contains "$OUT" "1/3" or "1/1" depending on the chosen policy, making the behavior explicit.

howarewoo (commented):

Replaces grep-based checkbox counting in `plan_progress()` with a fence-aware `awk` counter so example `- [ ]` boxes inside code fences no longer inflate plan progress. The fence open/close logic (CommonMark length rule, nested fences) is correct and well-tested; two non-blocking nits flag test-coverage gaps only.

---
**Status: APPROVED** — No blocking findings, 2 nit(s). See inline comments.
*Audited by woostack-review · Host: Claude Code · Provider: anthropic · Model: claude-sonnet-4-6 (workers) + claude-opus-4-8 (validators)*



skills/woostack-status/scripts/tests/test-status.sh
howarewoo: Fixed in `86c378e`. Added a plain triple-backtick fence fixture (2 done + 1 todo outside, 2 example boxes inside) asserting `2/3`/not `2/5`. Exercises the `flen=3` / `run>=flen` close path directly, so a `run==flen` regression now fails here even though the 4-backtick test still passes.

skills/woostack-status/scripts/tests/test-status.sh
howarewoo: Fixed in `86c378e`. Added an unclosed-fence fixture (1 done + 1 todo, then an opening ``` with no close, then 2 todos) asserting `1/2`/not `1/4` — pinning current behavior: `infence` stays 1 to EOF and trailing checkboxes are swallowed. Chose to characterize rather than change it: counting post-unclosed-fence boxes would need EOF-buffering of every pending-fence line, not worth it for malformed input. A comment documents the under-count footgun, and the assert makes any future awk change visible.

PR #407, opened 2026-06-17
Filter final review anchors

### LINKED ISSUE #375: Final findings line/path never re-resolved after validator passes → 422 / mis-anchored comments
## Summary

`resolve-diff-line` is run only in `prefetch.sh` and `merge-findings.sh` (on `raw_findings.json`, pre-validation). The validator passes (`validator.md` / `validator-prosecutor.md`) can rewrite a finding's `line`/`file`, and `intersect-findings.sh` blindly takes the **defender's** copy of `line`. Nothing re-validates the location before the posting stage, so a validator line error flows straight to `POST /reviews` and 422s (or, worse, posts an inline comment on the wrong line).

## Repro

- PR `howarewoo/TrioSens#1632`. Angle/prosecutor findings carried correct source lines (`chatgptAdapter.ts:300/306/308/309`, all resolvable).
- The defender pass emitted **diff.txt line positions** as `line` (e.g. `657`, `665` — the byte offsets of those code lines inside `diff.txt`, not source-file lines).
- `intersect-findings.sh` kept the defender's `line` (per its "other fields take the defender's copy" rule). Final `findings.json` lines `657`/`665` → `resolve-diff-line ... → null`.
- A direct POST with those lines would 422 (`Line could not be resolved`); the review only landed after manually re-anchoring each finding to the prosecutor's source lines.

## Root cause

- The diff-line resolver is a merge-stage gate. After merge, validators may mutate `line`/`file`, but there is no post-intersect re-resolution. The LLM slip (diff position vs source line) is the trigger; the missing deterministic guard is the bug.

## Proposed fix

- In `intersect-findings.sh` (or a thin post-step before posting), re-run `resolve-diff-line --file <path> --line <N>` on every finding in the final `findings.json`; drop (or `::warning::`-log) any that resolve to `null`. This makes the posted set provably anchorable regardless of what the validators wrote, and pairs naturally with the PR-file-set filter from the sibling incremental-marker issue.
- Optionally: when a finding resolves `null` but a sibling validator copy (prosecutor) resolves, prefer the resolvable line instead of dropping.

## Impact

On a self-PR these surface as a hard 422 (no review posted). On a normal PR a mis-anchored line can post an inline comment against unrelated code.

howarewoo:
## Goal
Fixes #375 by ensuring final woostack-review findings are postable after validator/intersection. The final `findings.json` is now filtered against the current PR file set and re-resolved against the diff before GitHub review posting can consume it.

## Summary
- Adds a final `intersect-findings.sh` anchor filter after classification in both defender-only and adversarial paths.
- Drops final findings outside the current PR file set or with unresolvable right-side diff anchors, and rewrites resolver output back to numeric `.line` values.
- Adds a regression test plus a scoped memory note for the final-payload anchor gotcha.

### Automated
- `bash skills/woostack-review/scripts/tests/test-intersect-final-anchors.sh`
- `bash skills/woostack-review/scripts/tests/test-intersect-farapart.sh`
- `bash skills/woostack-review/scripts/tests/test-intersect-nits.sh`
- `bash skills/woostack-review/scripts/tests/test-intersect-overlap.sh`
- `bash skills/woostack-review/scripts/tests/test-merge-findings-recovery.sh`
- `for t in skills/woostack-review/scripts/tests/test-*.sh; do bash "$t"; done`
- `bash -n skills/woostack-review/scripts/intersect-findings.sh && bash -n skills/woostack-review/scripts/tests/test-intersect-final-anchors.sh`
- `bash skills/woostack-doctor/scripts/doctor.sh --check` (0 errors; existing warnings: broad memory-overlap cluster and legacy plan-source link)

### Manual
- Reviewed the final diff for scope and confirmed only the fix plan, review script/test, and memory files are changed.

**Before merge**
- Confirm PR #407 review passes.

**After merge**
- No deploy or migration required.

Spec: .woostack/fixes/2026-06-17-final-finding-anchors.md


howarewoo:
* **#407** <a href="https://app.graphite.com/github/pr/howarewoo/woostack/407?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/howarewoo/woostack/407?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

skills/woostack-review/scripts/intersect-findings.sh
```diff
@@ -618,6 +745,7 @@ PY
 intersection_size="$(jq 'length' "$FINAL")"
```
howarewoo, line 745: **Nit: Cover adversarial final anchor filtering**

The new regression only runs defender-only mode with `disable_adversarial: true`, but the fix notes identify adversarial intersection as the path that can preserve a stale defender `file`/`line` after validator agreement. Without an adversarial-mode fixture, the test suite would not catch a regression where final filtering is removed or bypassed on the merged defender/prosecutor path.

Fix: Add an adversarial fixture with both `findings.defender.json` and `findings.prosecutor.json` present and `disable_adversarial` false. Include a stale defender anchor that still matches a prosecutor finding, then assert the final `findings.json` drops or canonicalizes it after intersection.

howarewoo (commented):
No blocking findings. One non-blocking nit is included inline.

STATUS: APPROVED

Note: GitHub does not allow approving your own PR, so this self-authored review is submitted as COMMENT.

Reviewed by woostack-review · Provider: openai · Model: gpt-5

skills/woostack-review/scripts/intersect-findings.sh
howarewoo: Fixed in b9ca2f4. Added test-intersect-final-anchors-adversarial.sh covering the merged defender/prosecutor path: a defender finding whose stale line (src/app.ts:99) is validated only by a pass-4 title match keeps the defender's stale anchor through the merge, and a stale-file finding (other.ts) survives intersection — the test asserts filter_final_anchors drops both and rewrites the one valid finding's string line to the canonical number. Also asserts validator-metrics.json mode==adversarial so a silent defender-only fallback can't mask a regression.

Issue #447, opened 2026-07-03
fix(review): angle workers load the full orchestrator SKILL.md (~14k tok each) via skill:// resolution


howarewoo:
## Description
Follow-up to #444 (and sibling of #445). Third leak in the same family: on hosts that resolve skills via a `skill://` reference (observed on omp; likely any host where the worker brief or subagent spawn carries the skill reference), each **angle worker** loads the **entire woostack-review `SKILL.md`** into its context — the orchestrator document (stages, prefetch, posting procedure, config schema, incremental mode, troubleshooting), none of which a worker needs or is allowed to act on.

Observed in omp session logs over 7 days (TrioSens consumer repo):
- Identical 57,116-char (~14k token) `skill://woostack-review` read present in every angle worker session: `ReviewBugs`, `ReviewAeo`, `ReviewTypes`, `ReviewDesign`, `ReviewArchitecture`, `ReviewSecurity`, `ReviewConventions`, … — **26 occurrences in one week**.
- It stacks on top of the shared-contract read (`_header.md`, fixed by #446) and the duplicated `rules.md` (#445) — three copies of overlapping orchestrator/rules context per worker.

A 13-angle review pays ~14k tok × 13 workers ≈ **~180k wasted tokens per review** from this leak alone.

## Steps to Reproduce
1. Run `/woostack-review` on a host whose subagent spawn resolves the invoking skill's `SKILL.md` into the worker context (omp reproduces; check other hosts' worker transcripts).
2. Inspect an angle worker's session transcript.
3. Full `SKILL.md` (frontmatter `name: woostack-review` … install line …) appears as a ~57k-char context block, alongside `_worker-header.md`/`_header.md` and the angle prompt.

## Expected Behavior
Angle workers receive only: the worker contract (`_worker-header.md` after #446), their single angle prompt, the diff/meta artifacts, and rules/memory context. The orchestrator `SKILL.md` never enters worker context.

## Actual Behavior
Every worker also carries the full ~14k-token orchestrator `SKILL.md`. Beyond the token cost, it hands workers instructions they are explicitly forbidden to execute (posting, prefetch, `$OUTDIR` lifecycle) — the Stage 3 brief then has to countermand them ("Sub-agents MUST NOT post comments, run prefetch.sh, …").

## Screenshot
N/A

## Additional Context
- Likely fix directions (host-dependent):
[...]

PR #462, opened 2026-07-09
fix: keep woostack-execute subagent briefs self-contained (no woostack-review misroute)

howarewoo:
## Goal

woostack-execute's dispatched reviewer/implementer subagents boot inside the consumer repo, auto-load `using-woostack`, and route "review"/"implement" intent into the full `skill://woostack-review` orchestrator (~14.7K tokens) — the wrong contract for a task-scoped reviewer, and pure token waste (16 such loads in one recent day of sessions, ~225K tokens). This extends the issue-#447 self-contained-brief guard from woostack-review's own angle workers to the execute subagent path.

## Summary

- Add a self-contained-brief guard inside the fenced brief of `implementer.md`, `spec-reviewer.md`, and `quality-reviewer.md`: do not load `skill://woostack-review` or route via `using-woostack`; follow only the brief.
- Record the boundary in `references/subagent-driver.md` (a "Self-contained briefs" note).
- Add `scripts/tests/test-subagent-brief-skill-scope.sh` + `run-tests.sh` (woostack-execute had no test harness), mirroring woostack-review's `test-worker-brief-skill-scope.sh`.
- Generalize the `review-workers-avoid-skill-scope` memory note to cover both mechanisms (skill-scoped propagation + using-woostack routing) and both skills.
- Covers woostack-execute-overnight too — it reuses these prompts and ships no duplicates.

### Automated

- [ ] `bash skills/woostack-execute/scripts/tests/run-tests.sh` — 8 passed, 0 failed (was 8-failed red before the guard).
- [ ] `bash skills/woostack-review/scripts/tests/test-worker-brief-skill-scope.sh` — pass (no regression to the #447 guard it extends).
- [ ] `bash skills/woostack-doctor/scripts/doctor.sh .` — 0 errors (memory note clean after index regen).

### Manual

**Before merge**

- [ ] Confirm the guard sentence sits inside the fenced brief block each subagent receives, in all three prompts.
- [ ] Read `subagent-driver.md`'s "Self-contained briefs" note for the two-mechanism rationale.

Spec: .woostack/fixes/2026-07-09-execute-subagent-brief-skill-scope.md


howarewoo:
* **#462** <a href="https://app.graphite.com/github/pr/howarewoo/woostack/462?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/howarewoo/woostack/462?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

skills/woostack-execute/scripts/tests/test-subagent-brief-skill-scope.sh
```diff
@@ -0,0 +1,43 @@
...
+    echo "    $(basename "$file") does not match [$regex]"
+  fi
+}
+
+PROMPTS="$ROOT/skills/woostack-execute/prompts"
+DRIVER="$ROOT/skills/woostack-execute/references/subagent-driver.md"
+
+# Each dispatched brief, plus the driver that records the dispatch contract, must declare
+# itself self-contained AND forbid loading the woostack-review orchestrator skill.
+for f in "$PROMPTS/implementer.md" "$PROMPTS/spec-reviewer.md" "$PROMPTS/quality-reviewer.md" "$DRIVER"; do
```
howarewoo: **Nit: Test reads whole file, not the dispatched fenced brief**

Both assertions read the entire prompt file (`cat "$file"` and `tr '\n' ' ' < "$file"`), but only the fenced block is sent to the subagent: implementer.md line 8 says "send the fenced block below as the subagent prompt", and each of the three prompts wraps its brief in a ```` fence with a controller-facing preamble above it. A guard sentence placed in that preamble (outside the fence) would satisfy both assertions yet never reach the subagent, so the reintroduced bug would pass green. This is the exact placement invariant the fix depends on, which is why the fix doc falls back to a manual Step-3 grep to confirm it, work the automated guard should be doing.

Fix: For the three prompt files, scope the assertions to the fenced brief region (the text between the opening and closing ```` markers) before matching, so a guard added to the controller preamble fails. subagent-driver.md is read whole by the controller, so its whole-file check is fine.

skills/woostack-execute/scripts/tests/test-subagent-brief-skill-scope.sh
```diff
@@ -0,0 +1,43 @@
...
+  assert_file_contains "$f" "self-contained" \
+    "$(basename "$f") must declare the subagent brief self-contained"
+  assert_file_matches "$f" "(do not|never).*skill://woostack-review" \
```
howarewoo: **Nit: Greedy regex lets any earlier negation pass guard**

The negation-binding check flattens the whole file to one line and matches `(do not|never).*skill://woostack-review`. The unbounded greedy `.*` only requires *some* `do not`/`never` to appear anywhere before the token anywhere in the file; it never ties the negation to the guard sentence itself. These prompts contain several unrelated negations (`do not git-commit`, `Never trade away`, `do not create`), so a malformed guard whose own negation is dropped but whose `skill://woostack-review` token survives still passes green. Verified concretely: a brief reading `...do not git-commit... you may load skill://woostack-review freely` matches, so the exact regression this test exists to catch can slip through. Memory `grep-assertion` guidance warns to pin the contract, not a near-match.

Fix: Bound the gap so the negation and token must sit in the same sentence, e.g. replace the unbounded `.*` with `[^.]*` (no sentence-ending period between). Confirmed all four current files still match.

```suggestion
  assert_file_matches "$f" "(do not|never)[^.]*skill://woostack-review" \
```

howarewoo: **Nit: Guard's using-woostack routing clause is unasserted**

The guard sentence forbids two distinct load paths, `skill://woostack-review` AND `using-woostack` command routing, and the cross-PR memory records the using-woostack routing path as the *primary* mechanism this self-contained brief fixes ("for (2) the self-contained brief IS the fix, because the profile is already plain/general-purpose and still routes without it"). The test only pins the `skill://woostack-review` token; nothing asserts the guard still addresses the using-woostack routing path. A future edit that drops the `or using-woostack command routing` clause would leave the fix's primary documented failure mode unguarded while the test stays green.

Fix: Add a companion assertion that the guard also names the using-woostack routing boundary, e.g. match `(do not|never)[^.]*using-woostack` on the same flattened content, so both documented load paths are pinned.

howarewoo (commented):

Extends the issue-#447 self-contained-brief guard from woostack-review's own angle workers to woostack-execute's dispatched subagents — adding a "do NOT load `skill://woostack-review` or route via `using-woostack`" sentence inside the fenced brief of `implementer.md`, `spec-reviewer.md`, and `quality-reviewer.md` (plus a rationale note in `subagent-driver.md`), a new bash regression test (`run-tests.sh` + `test-subagent-brief-skill-scope.sh`), and a widened `review-workers-avoid-skill-scope` memory note. The change is well-scoped and correct — the guard lands inside each dispatched fence, and bugs/security/simplify/conventions/docs all came back clean; the three findings below are non-blocking `tests`-angle nits about the new regression test's precision (it reads the whole file rather than the dispatched fenced brief, and its greedy regex + missing `using-woostack` assertion could be tightened).

---
**Status: APPROVED** — No blocking findings, 3 nit(s). See inline comments.
*Audited by woostack-review · Host: oh-my-pi · Provider: anthropic · Model: claude-opus-4-8*



skills/woostack-execute/scripts/tests/test-subagent-brief-skill-scope.sh
howarewoo: Fixed in 3cf5c1e. The three prompt assertions now run against the dispatched four-backtick fenced brief (extracted by a new `fenced_brief` helper), not the whole file — so a guard placed in the controller-facing preamble outside the fence now fails. `subagent-driver.md` stays a whole-file check since it has no dispatched fence. Verified red against a guard moved out of the fence.

skills/woostack-execute/scripts/tests/test-subagent-brief-skill-scope.sh
howarewoo: Fixed in 3cf5c1e. The /Users/adamwoo/.claude/skills/woostack-review/SKILL.md check now binds the negation to the guard sentence with `(do not|never)[^.]*/Users/adamwoo/.claude/skills/woostack-review/SKILL.md` instead of the greedy `.*`, so an unrelated earlier `do not`/`never` in the brief can no longer satisfy the match when the guard's own negation is dropped. Verified: the greedy form false-passes that regression, the `[^.]*` form catches it.

howarewoo: Fixed in 3cf5c1e. Added an assertion for the guard's using-woostack clause, matching the backtick-wrapped `using-woostack` token. It matches the backtick-wrapped form specifically so the unrelated `../../using-woostack/...` doc link in the briefs cannot false-satisfy it; `[^.]*` can't be reused here because the `SKILL.md` inline-code period sits between the negation and `using-woostack`.

howarewoo: Correction — my previous reply had the `skill://woostack-review` URI auto-expanded to a local filesystem path by the harness tooling. The intended text:

The negation-binding check now uses `(do not|never)[^.]*skill://woostack-review` (bounded `[^.]*`) instead of the greedy `(do not|never).*skill://woostack-review`, so an unrelated earlier `do not`/`never` in the brief can no longer satisfy the match when the guard's own negation is dropped. Verified: the greedy form false-passes that regression, the bounded form catches it. Landed in 3cf5c1e.

PR #466, opened 2026-07-09
feat(init): add gen-omp-agents generator and gitignore

howarewoo:
## Goal

Introduce the single generation authority for baking woostack config models into omp agent definitions, ensuring generated files are ignored by git.

## Summary

- Create `skills/woostack-init/scripts/gen-omp-agents.sh` to translate `.woostack/config.json` `models.<tier>` config to `.omp/agents/woostack-*.md` defs.
- Create unit test suite `skills/woostack-init/scripts/tests/test-gen-omp-agents.sh` verifying all AC1-AC4 criteria (string/object configuration, unset tiers, safety checks, idempotence, worktrees).
- Configure gitignore rules in `.gitignore` and `.omp/agents/.gitignore` to ignore generated agent files.

### Automated

- [x] `bash skills/woostack-init/scripts/tests/test-gen-omp-agents.sh` (19 passed, 0 failed)
- [x] `bash -n skills/woostack-init/scripts/gen-omp-agents.sh` (syntax OK)
- [x] `git check-ignore` verifies generated agents are ignored, custom ones kept (0/1 exit codes correct)

### Manual

**Before merge**

- [x] Spawned `agent: woostack-fast` via `task` tool to confirm the markdown system prompt loads and inherits standard tools successfully.

Spec: .woostack/specs/2026-07-09-omp-model-tiers.md

### DIFF
--- a/skills/woostack-init/scripts/gen-omp-agents.sh
+++ b/skills/woostack-init/scripts/gen-omp-agents.sh
@@ -53,11 +53,7 @@ safe_slug() {
 
 render_tier() {
   local tier="$1" leaf ltype model effort tl f
-  if [ -f "$CFG" ]; then
-    leaf="$(jq -c --arg t "$tier" '(.models // {})[$t] // null' "$CFG" 2>/dev/null || echo null)"
-  else
-    leaf="null"
-  fi
+  leaf="$(jq -c --arg t "$tier" '(.models // {})[$t] // null' "$CFG" 2>/dev/null || echo null)"
   ltype="$(printf '%s' "$leaf" | jq -r 'type' 2>/dev/null || echo null)"
   model=""; effort=""
   case "$ltype" in
@@ -82,7 +78,7 @@ render_tier() {
   if [ -n "$effort" ] && valid_effort "$effort"; then
     tl="$effort"
   else
-    [ -n "$effort" ] && echo "gen-omp-agents.sh: $tier: effort '$effort' not in enum; using tier default" >&2
+    [ -n "$effort" ] && echo "gen-omp-agents.sh: $tier: effort not in enum; using tier default" >&2
     tl="$(default_effort "$tier")"
   fi
 
--- a/skills/woostack-init/scripts/gen-omp-agents.sh
+++ b/skills/woostack-init/scripts/gen-omp-agents.sh
@@ -31,14 +31,25 @@ mkdir -p "$OUT_DIR"
 # gitignore generated defs (scoped: woostack-*.md only; a consumer's own custom.md stays tracked)
 ignore="$OUT_DIR/.gitignore"
 if ! [ -f "$ignore" ] || ! grep -qxF 'woostack-*.md' "$ignore" 2>/dev/null; then
+  # if the file exists without a trailing newline, add one first so the pattern
+  # can't glue onto a consumer's unterminated last entry (custom-entrywoostack-*.md)
+  if [ -s "$ignore" ] && [ -n "$(tail -c1 "$ignore")" ]; then printf '\n' >> "$ignore"; fi
   printf '%s\n' 'woostack-*.md' >> "$ignore"
 fi
 
 # omp thinkingLevel enum (verified: omp models.md). woostack effort maps 1:1 (+ off).
 valid_effort() { case "$1" in off|minimal|low|medium|high|xhigh) return 0 ;; *) return 1 ;; esac; }
 default_effort() { case "$1" in fast) echo low ;; standard) echo medium ;; deep) echo xhigh ;; esac; }
-# model slug charset guard -> no YAML metachars/newlines -> no frontmatter injection
-safe_slug() { printf '%s' "$1" | grep -qE '^[A-Za-z0-9][A-Za-z0-9._/:+-]*$'; }
+# model slug charset guard -> no YAML metachars/newlines -> no frontmatter injection.
+# Whole-string case match (bash globs span newlines) so a multi-line value can't
+# slip a clean first line past a per-line grep and inject frontmatter/body.
+safe_slug() {
+  case "$1" in
+    ''|*[!A-Za-z0-9._/:+-]*) return 1 ;;
+    [A-Za-z0-9]*) return 0 ;;
+    *) return 1 ;;
+  esac
+}
 
 render_tier() {
   local tier="$1" leaf ltype model effort tl f
--- a/skills/woostack-init/scripts/gen-omp-agents.sh
+++ b/skills/woostack-init/scripts/gen-omp-agents.sh
@@ -1,0 +1,93 @@
+#!/usr/bin/env bash
+# gen-omp-agents.sh - bake .woostack/config.json models.<tier> into omp agent-defs
+#   <primary-root>/.omp/agents/woostack-{fast,standard,deep}.md
+# Single generation authority; callers: woostack-init (scaffold), woostack-doctor
+# (--fix), woostack-execute (safety-net). Idempotent (overwrite). Best-effort + loud:
+# a malformed/unsafe leaf -> that tier unset (thinkingLevel-only) + stderr warn, exit 0.
+set -uo pipefail
+
+# --- primary-root resolution (worktree contract; mirrors resolve-root.sh) ---
+# Precedence: WOOSTACK_ROOT > git-common-dir parent (primary root from any worktree) > pwd.
+resolve_root() {
+  if [ -n "${WOOSTACK_ROOT:-}" ]; then
+    ( cd "$WOOSTACK_ROOT" 2>/dev/null && pwd -P ) || printf '%s\n' "$WOOSTACK_ROOT"
+    return
+  fi
+  local cg
+  cg="$(git rev-parse --git-common-dir 2>/dev/null || true)"
+  if [ -n "$cg" ]; then
+    ( cd "$cg/.." 2>/dev/null && pwd -P ) && return
+  fi
+  pwd -P
+}
+
+ROOT="$(resolve_root)"
+CFG="$ROOT/.woostack/config.json"
+OUT_DIR="${WOO_OMP_AGENTS_DIR:-$ROOT/.omp/agents}"
+
+command -v jq >/dev/null 2>&1 || { echo "gen-omp-agents.sh: jq not found; skipping" >&2; exit 0; }
+mkdir -p "$OUT_DIR"
+
+# gitignore generated defs (scoped: woostack-*.md only; a consumer's own custom.md stays tracked)
+ignore="$OUT_DIR/.gitignore"
+if ! [ -f "$ignore" ] || ! grep -qxF 'woostack-*.md' "$ignore" 2>/dev/null; then
+  printf '%s\n' 'woostack-*.md' >> "$ignore"
+fi
+
+# omp thinkingLevel enum (verified: omp models.md). woostack effort maps 1:1 (+ off).
+valid_effort() { case "$1" in off|minimal|low|medium|high|xhigh) return 0 ;; *) return 1 ;; esac; }
+default_effort() { case "$1" in fast) echo low ;; standard) echo medium ;; deep) echo xhigh ;; esac; }
+# model slug charset guard -> no YAML metachars/newlines -> no frontmatter injection
+safe_slug() { printf '%s' "$1" | grep -qE '^[A-Za-z0-9][A-Za-z0-9._/:+-]*$'; }
+
+render_tier() {
+  local tier="$1" leaf ltype model effort tl f
+  if [ -f "$CFG" ]; then
+    leaf="$(jq -c --arg t "$tier" '(.models // {})[$t] // null' "$CFG" 2>/dev/null || echo null)"
+  else
+    leaf="null"
+  fi
+  ltype="$(printf '%s' "$leaf" | jq -r 'type' 2>/dev/null || echo null)"
+  model=""; effort=""
+  case "$ltype" in
+    string)
+      model="$(printf '%s' "$leaf" | jq -r '.')"
+      [ -n "$model" ] || echo "gen-omp-agents.sh: $tier: empty string leaf; tier unset" >&2
+      ;;
+    object)
+      model="$(printf '%s' "$leaf" | jq -r '.model // ""')"
+      effort="$(printf '%s' "$leaf" | jq -r '.effort // ""')"
+      [ -n "$model" ] || echo "gen-omp-agents.sh: $tier: object leaf missing .model; tier unset" >&2
+      ;;
+    null) : ;;
+    *) echo "gen-omp-agents.sh: $tier: malformed leaf ($ltype); tier unset" >&2 ;;
+  esac
+
+  if [ -n "$model" ] && ! safe_slug "$model"; then
+    echo "gen-omp-agents.sh: $tier: unsafe model slug; tier unset" >&2
+    model=""
+  fi
+
+  if [ -n "$effort" ] && valid_effort "$effort"; then
+    tl="$effort"
+  else
+    [ -n "$effort" ] && echo "gen-omp-agents.sh: $tier: effort '$effort' not in enum; using tier default" >&2
+    tl="$(default_effort "$tier")"
+  fi
+
+  f="$OUT_DIR/woostack-$tier.md"
+  {
+    printf -- '---\n'
+    printf 'name: woostack-%s\n' "$tier"
+    printf 'description: woostack %s-tier general-purpose worker (generated from .woostack/config.json; edits are overwritten).\n' "$tier"
+    [ -n "$model" ] && printf 'model: "%s"\n' "$model"
+    printf 'thinkingLevel: %s\n' "$tl"
+    printf -- '---\n\n'
+    printf 'You are a general-purpose woostack worker running at the %s tier. The task you receive on\n' "$tier"
+    printf 'the role / context / assignment fields is authoritative - do exactly what it specifies; it\n'
+    printf 'carries the full context you need. Do not load skill://woostack-review or route through\n'
+    printf 'using-woostack.\n'
+  } > "$f"
+}
+
+for tier in fast standard deep; do render_tier "$tier"; done
--- a/skills/woostack-init/scripts/tests/test-gen-omp-agents.sh
+++ b/skills/woostack-init/scripts/tests/test-gen-omp-agents.sh
@@ -38,16 +38,44 @@ assert_eq "$([ -f "$f" ] && echo y)" "y" "AC1 unset: standard def still written"
 assert_not_contains "$(cat "$f")" 'model:' "AC1 unset: no model line"
 assert_contains "$(cat "$f")" 'thinkingLevel: medium' "AC1 unset: standard default=medium"
 
+# --- AC1 missing config -> same unset-tier fallback ---
+r="$(mktemp -d)"; ( cd "$r" && git init -q )
+WOOSTACK_ROOT="$r" bash "$GEN"
+f="$r/.omp/agents/woostack-standard.md"
+assert_eq "$([ -f "$f" ] && echo y)" "y" "AC1 missing config: standard def still written"
+assert_not_contains "$(cat "$f")" 'model:' "AC1 missing config: no model line"
+assert_contains "$(cat "$f")" 'thinkingLevel: medium' "AC1 missing config: standard default=medium"
+
+# --- AC1 error: empty-string leaf -> tier unset, def still written (default thinkingLevel) ---
+r="$(mktemp -d)"; ( cd "$r" && git init -q )
+mkcfg "$r" '{ "fast": "" }'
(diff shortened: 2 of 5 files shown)


howarewoo:
* **#470** <a href="https://app.graphite.com/github/pr/howarewoo/woostack/470?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#469** <a href="https://app.graphite.com/github/pr/howarewoo/woostack/469?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#468** <a href="https://app.graphite.com/github/pr/howarewoo/woostack/468?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#467** <a href="https://app.graphite.com/github/pr/howarewoo/woostack/467?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#466** <a href="https://app.graphite.com/github/pr/howarewoo/woostack/466?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/howarewoo/woostack/466?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

skills/woostack-init/scripts/gen-omp-agents.sh
```diff
@@ -0,0 +1,93 @@
...
+CFG="$ROOT/.woostack/config.json"
+OUT_DIR="${WOO_OMP_AGENTS_DIR:-$ROOT/.omp/agents}"
+
+command -v jq >/dev/null 2>&1 || { echo "gen-omp-agents.sh: jq not found; skipping" >&2; exit 0; }
+mkdir -p "$OUT_DIR"
+
+# gitignore generated defs (scoped: woostack-*.md only; a consumer's own custom.md stays tracked)
+ignore="$OUT_DIR/.gitignore"
+if ! [ -f "$ignore" ] || ! grep -qxF 'woostack-*.md' "$ignore" 2>/dev/null; then
+  printf '%s\n' 'woostack-*.md' >> "$ignore"
```
howarewoo, line 37: **Nit: gitignore append can glue onto an unterminated last line**

When a consumer-authored `$OUT_DIR/.gitignore` exists but its final line lacks a trailing newline, `printf '%s\n' 'woostack-*.md' >> "$ignore"` glues the pattern onto that last entry (e.g. `custom-entrywoostack-*.md`), corrupting both patterns: the consumer's entry stops matching and generated defs stay unignored. The next run's `grep -qxF` still misses, so the line is appended again — converging on run two, but the corrupted glued entry persists. The generator's own writes are newline-terminated, so this only bites on pre-existing consumer files — and `.omp/agents/.gitignore` is explicitly a consumer-owned surface per the line-31 comment ("a consumer's own custom.md stays tracked").

Fix: Before appending, emit a newline when the existing file's last byte is not one.

```suggestion
  [ -s "$ignore" ] && [ -n "$(tail -c1 "$ignore")" ] && printf '\n' >> "$ignore"
  printf '%s\n' 'woostack-*.md' >> "$ignore"
```

<sub>— <strong>LOW · NIT</strong> · flagged by the <code>bugs</code> agent</sub>

skills/woostack-init/scripts/gen-omp-agents.sh
```diff
@@ -0,0 +1,93 @@
...
+fi
+
+# omp thinkingLevel enum (verified: omp models.md). woostack effort maps 1:1 (+ off).
+valid_effort() { case "$1" in off|minimal|low|medium|high|xhigh) return 0 ;; *) return 1 ;; esac; }
+default_effort() { case "$1" in fast) echo low ;; standard) echo medium ;; deep) echo xhigh ;; esac; }
+# model slug charset guard -> no YAML metachars/newlines -> no frontmatter injection
+safe_slug() { printf '%s' "$1" | grep -qE '^[A-Za-z0-9][A-Za-z0-9._/:+-]*$'; }
```
howarewoo: **safe_slug newline bypass allows frontmatter injection**

safe_slug pipes the value to `grep -qE '^...$'`, but grep anchors per physical line: for a multi-line value it succeeds when ANY single line matches the slug charset. A config leaf like `"fast": "good/slug\nmalicious: injected"` therefore passes the guard, and `printf 'model: "%s"\n'` emits the raw newline into the generated def. Verified live: the output frontmatter contains `model: "good/slug` followed by a live `malicious: injected` key. A payload of the form `x/y\n---\n<text>` can likewise terminate the frontmatter early and inject arbitrary agent-def body (system prompt) content. This defeats the guard's stated contract on line 40 ("no YAML metachars/newlines -> no frontmatter injection"). The existing AC1 injection test only uses a payload whose first line contains a `"` (so no line matches), and does not catch this bypass.

Fix: Reject any value containing a newline before the per-line grep (bash pattern match on `$'\n'`), and extend the AC1 injection test with a first-line-valid payload such as `"good/slug\nmalicious: injected"` to pin the fix.

```suggestion
safe_slug() { case "$1" in *$'\n'*) return 1 ;; esac; printf '%s' "$1" | grep -qE '^[A-Za-z0-9][A-Za-z0-9._/:+-]*$'; }
```

skills/woostack-init/scripts/tests/test-gen-omp-agents.sh
```diff
@@ -0,0 +1,92 @@
...
+mkcfg "$r" '{ "standard": "openai/gpt-5.5" }'
+WOOSTACK_ROOT="$r" bash "$GEN"; a="$(cat "$r/.omp/agents/woostack-standard.md")"
+WOOSTACK_ROOT="$r" bash "$GEN"; b="$(cat "$r/.omp/agents/woostack-standard.md")"
+assert_eq "$a" "$b" "AC2 idempotent: identical output"
+
+# --- AC2 edge: run from a worktree cwd, no WOOSTACK_ROOT -> primary root ---
+r="$(mktemp -d)"; ( cd "$r" && git init -q && git commit -q --allow-empty -m init )
+mkcfg "$r" '{ "fast": "p/q" }'
+wt="$(mktemp -d)"; ( cd "$r" && git worktree add -q "$wt" -b wt-branch )
+( cd "$wt" && bash "$GEN" )
```
howarewoo: **Nit: Worktree test inherits WOOSTACK_ROOT from environment**

The AC2 comment on line 75 says "no WOOSTACK_ROOT", but `( cd "$wt" && bash "$GEN" )` inherits the caller's environment. Woostack tooling exports WOOSTACK_ROOT in several flows; when it leaks into a test run, the generator under test resolves that root first (its highest-precedence branch), writes `woostack-*.md` defs into the user's real repo instead of the temp fixture, and the primary-root assertion on line 79 fails spuriously. Every other case in this suite pins WOOSTACK_ROOT explicitly; this is the only case that relies on the variable being unset without enforcing it.

Fix: Invoke the generator with the variable explicitly cleared so the test exercises the git-common-dir resolution path deterministically.

```suggestion
( cd "$wt" && env -u WOOSTACK_ROOT bash "$GEN" )
```

<sub>— <strong>LOW · NIT</strong> · flagged by the <code>bugs</code> agent</sub>

skills/woostack-init/scripts/gen-omp-agents.sh
howarewoo, line 37: **Nit: Gitignore append-once and preservation unpinned by tests**

This append writes into a consumer-owned file (`$OUT_DIR/.gitignore`), guarded by `grep -qxF` so it appends exactly once and preserves existing entries. Neither property is pinned: AC4 always starts from a fresh temp dir (so the file never pre-exists), and the AC2 idempotency case compares only the generated def file, not the .gitignore. A regression flipping `>>` to `>` (clobbering a consumer's own ignore entries) or dropping the grep guard (duplicating `woostack-*.md` on every run of this thrice-called generator) passes the current 19-assertion suite untouched.

Fix: Add a test case that pre-seeds `$r/.omp/agents/.gitignore` with a custom entry (e.g. `custom-ignored.md`), runs the generator twice, then asserts the custom entry survives and `grep -cxF 'woostack-*.md'` reports exactly 1.

skills/woostack-init/scripts/tests/test-gen-omp-agents.sh
```diff
@@ -0,0 +1,92 @@
...
+assert_eq "$([ -f "$r/.omp/agents/woostack-fast.md" ] && echo y)" "y" "AC2 worktree: def at primary root"
+assert_eq "$([ -f "$wt/.omp/agents/woostack-fast.md" ] && echo n || echo n)" "n" "AC2 worktree: not in worktree tree"
```
howarewoo: **Nit: Worktree assertion is tautological, verifies nothing**

`$([ -f "$wt/.omp/agents/woostack-fast.md" ] && echo n || echo n)` echoes `n` on both branches, so `assert_eq ... "n"` can never fail — the 'AC2 worktree: not in worktree tree' check is dead weight as written and the negative half of the worktree contract is untested. Either the branch should echo `y`/`n` so the assertion actually reflects file existence, or the line should be deleted rather than kept as a check that always passes.

Fix: Echo `y` when the file exists so the assertion genuinely fails if the def lands inside the worktree.

```suggestion
assert_eq "$([ -f "$wt/.omp/agents/woostack-fast.md" ] && echo y || echo n)" "n" "AC2 worktree: not in worktree tree"
```

<sub>— <strong>MEDIUM · NIT</strong> · flagged by the <code>simplify</code> agent</sub>

howarewoo (commented):

Reviewed the OMP agents generator increment (`gen-omp-agents.sh`, `.gitignore`, and its test). One blocking correctness/security defect: the `safe_slug` charset guard is bypassable by a newline-bearing config value, allowing YAML frontmatter injection into generated agent defs. Four nits on test hermeticity, coverage, and gitignore handling.

---
**Status: CHANGES REQUESTED** — 1 blocking + 0 non-blocking finding(s) (1 HIGH, 0 MEDIUM, 0 LOW) + 4 nit(s). See inline comments.
*Audited by woostack-review · Host: oh-my-pi · Provider: anthropic · Model: claude-fable-5*



skills/woostack-init/scripts/gen-omp-agents.sh
howarewoo: Fixed in `0d5c7ff`. safe_slug now uses a whole-string bash `case` guard (globs span newlines), so a multi-line value can no longer slip a clean first line past the per-line charset check. Added a test with a clean-first-line multi-line slug that pins the bypass closed.

skills/woostack-init/scripts/gen-omp-agents.sh
howarewoo: Fixed in `0d5c7ff`. The generator now adds a trailing newline before appending when the consumer .gitignore lacks one, so the pattern can't glue onto an unterminated last entry.

skills/woostack-init/scripts/tests/test-gen-omp-agents.sh
howarewoo: Fixed in `0d5c7ff`. The worktree case now runs with `env -u WOOSTACK_ROOT`, so the test no longer inherits an ambient WOOSTACK_ROOT.

skills/woostack-init/scripts/gen-omp-agents.sh
howarewoo: Fixed in `0d5c7ff`. Added AC4b: pins append-exactly-once across two runs and preservation of a consumer entry seeded without a trailing newline.

skills/woostack-init/scripts/tests/test-gen-omp-agents.sh
howarewoo: Fixed in `0d5c7ff`. The assertion now uses `echo y || echo n` so it actually reflects presence and verifies the def is absent from the worktree tree.

skills/woostack-init/scripts/gen-omp-agents.sh
```diff
@@ -0,0 +1,104 @@
...
+    leaf="$(jq -c --arg t "$tier" '(.models // {})[$t] // null' "$CFG" 2>/dev/null || echo null)"
+  else
+    leaf="null"
+  fi
+  ltype="$(printf '%s' "$leaf" | jq -r 'type' 2>/dev/null || echo null)"
+  model=""; effort=""
+  case "$ltype" in
+    string)
+      model="$(printf '%s' "$leaf" | jq -r '.')"
+      [ -n "$model" ] || echo "gen-omp-agents.sh: $tier: empty string leaf; tier unset" >&2
```
howarewoo, line 62: **Nit: Empty-string / missing-.model leaf branches untested**

render_tier has two distinct warn-and-unset error branches with no coverage: a string leaf that is empty (e.g. `{"fast":""}`, this line) and an object leaf missing `.model` (e.g. `{"deep":{"effort":"high"}}`, line 71). The suite's only unset-via-error case is the malformed number leaf (`{"fast":42}`), which exits through the `*)` branch and never reaches these `[ -n "$model" ]` guards, so a regression that emitted `model: ""` here would ship green. The malformed and injection cases also assert only the absence of a `model:` line; no error-path case positively asserts the def is still written with a tier-default `thinkingLevel:`, so a change that bailed before writing the file would pass vacuously (cat of a missing file is empty).

Fix: Add two cases mirroring the malformed-leaf test: config `{"fast":""}` and `{"deep":{"effort":"high"}}`, each asserting both `assert_not_contains ... 'model:'` AND `assert_contains ... 'thinkingLevel:'` (tier default). That pins both empty-string and missing-.model branches and verifies the graceful-degradation contract (tier unset but def still emitted) instead of relying on a vacuous absence check.

howarewoo (commented):

Re-review (round 2, full) after the fixes. The round-1 blocking safe_slug frontmatter-injection and the tautological/non-hermetic test issues are all confirmed resolved (bugs, security, simplify, conventions, docs clean). One remaining LOW nit: the empty-string and missing-`.model` error branches in `render_tier` are untested.

---
**Status: APPROVED** — No blocking findings, 1 nit(s). See inline comments.
*Audited by woostack-review · Host: oh-my-pi · Provider: anthropic · Model: claude-fable-5*



skills/woostack-init/scripts/gen-omp-agents.sh
howarewoo: Fixed in `0b49a48`. Added two tests: an empty-string leaf and an object leaf missing `.model`, each asserting the def is still emitted with a tier-default (or configured) thinkingLevel and no model line.

skills/woostack-init/scripts/gen-omp-agents.sh
```diff
@@ -0,0 +1,104 @@
...
+
+  if [ -n "$model" ] && ! safe_slug "$model"; then
+    echo "gen-omp-agents.sh: $tier: unsafe model slug; tier unset" >&2
+    model=""
+  fi
+
+  if [ -n "$effort" ] && valid_effort "$effort"; then
+    tl="$effort"
+  else
+    [ -n "$effort" ] && echo "gen-omp-agents.sh: $tier: effort '$effort' not in enum; using tier default" >&2
```
howarewoo: **Nit: Untrusted effort value echoed raw to stderr**

The effort branch echoes the raw, untrusted `$effort` value straight to stderr on the enum-miss path. `.woostack/config.json` is the same hostile-config trust boundary the `safe_slug` model guard defends against: a repo you run woostack-init / doctor --fix / execute against can set `models.<tier>.effort` to a string carrying terminal escape sequences or forged log lines, which are emitted verbatim to the operator's terminal. Every sibling warning is careful the other way -- the model warning on line 78 omits its untrusted value, and the malformed-leaf warning only prints jq's enum-bounded `$ltype` -- so this effort path is the one place attacker-controlled config content reaches the terminal, an inconsistency with the script's own hardening.

Fix: Drop the raw value from the message (mirroring the model warning on line 78), or sanitize it to a printable-only form before echoing, so attacker-controlled config bytes cannot reach the terminal.

```suggestion
    [ -n "$effort" ] && echo "gen-omp-agents.sh: $tier: effort not in enum; using tier default" >&2
```

<sub>— <strong>LOW · NIT</strong> · flagged by the <code>security</code> agent</sub>

skills/woostack-init/scripts/gen-omp-agents.sh
```diff
@@ -0,0 +1,104 @@
...
+  case "$1" in
+    ''|*[!A-Za-z0-9._/:+-]*) return 1 ;;
+    [A-Za-z0-9]*) return 0 ;;
+    *) return 1 ;;
+  esac
+}
+
+render_tier() {
+  local tier="$1" leaf ltype model effort tl f
+  if [ -f "$CFG" ]; then
```
howarewoo: **Nit: Redundant config-file guard duplicates jq null fallback**

In `render_tier`, the `if [ -f "$CFG" ]; then ... else leaf="null"; fi` wrapper (lines 56-60) reproduces a fallback the jq call already provides. The jq line ends in `2>/dev/null || echo null`, so when `$CFG` is absent jq exits non-zero and `leaf` becomes `null` anyway. There is no pipe on this line and `set -e` is not active (only `set -uo pipefail`), so the wrapped and unwrapped forms are behaviorally identical; the explicit `-f` test plus its `else` branch are four lines that buy nothing.

Fix: Delete the `if [ -f "$CFG" ]` / `else leaf="null"` / `fi` wrapper and keep only the assignment `leaf="$(jq -c --arg t "$tier" '(.models // {})[$t] // null' "$CFG" 2>/dev/null || echo null)"`, which already yields `null` when the config file does not exist.

<sub>— <strong>LOW · NIT</strong> · flagged by the <code>simplify</code> agent</sub>

skills/woostack-init/scripts/gen-omp-agents.sh
```diff
@@ -0,0 +1,104 @@
...
+# omp thinkingLevel enum (verified: omp models.md). woostack effort maps 1:1 (+ off).
+valid_effort() { case "$1" in off|minimal|low|medium|high|xhigh) return 0 ;; *) return 1 ;; esac; }
+default_effort() { case "$1" in fast) echo low ;; standard) echo medium ;; deep) echo xhigh ;; esac; }
+# model slug charset guard -> no YAML metachars/newlines -> no frontmatter injection.
+# Whole-string case match (bash globs span newlines) so a multi-line value can't
+# slip a clean first line past a per-line grep and inject frontmatter/body.
+safe_slug() {
+  case "$1" in
+    ''|*[!A-Za-z0-9._/:+-]*) return 1 ;;
+    [A-Za-z0-9]*) return 0 ;;
```
howarewoo, line 49: **Nit: safe_slug leading-char reject branch untested**

safe_slug has three branches: reject on empty/disallowed-char (line 47), accept on an alphanumeric-leading slug (line 48), and reject a slug whose chars are all in the allowed set but that starts with a non-alphanumeric char (line 49, the `*) return 1` case: leading `-`, `/`, `.`, `:`, or `+`). The suite pins branch 1 via the newline-injection case and branch 2 via the valid slugs (`google/gemini-3-5-flash`, `x/y`, `p/q`), but no test exercises branch 3 — e.g. a value like `-oProxyCommand=x` or `/etc/passwd`. Since this guard is the anti-injection boundary, a refactor that loosened case 3 to accept would let a flag-like or absolute-path-like slug land unguarded in the generated frontmatter `model:` line, and no test would catch the regression.

Fix: Add an assertion beside the existing injection case: configure a charset-valid but leading-non-alphanumeric model (e.g. `{ "fast": "-oProxyCommand=x" }` or `{ "fast": "/etc/passwd" }`), run the generator, and assert the generated woostack-fast.md contains no `model:` line (tier unset) — pinning safe_slug's third branch.

howarewoo (commented):

Re-review (round 3, full) at HEAD 0b49a48. The round-2 test-coverage nit is confirmed fixed. No blocking findings. Three residual LOW nits remain (all optional, non-blocking): a raw `$effort` echo to stderr (terminal-escape/log-injection on the config trust boundary), a redundant config-file guard that duplicates the jq `|| echo null` fallback, and the `safe_slug` leading-char reject branch (case 3) being untested. This is the round-3 (max_rounds) cap; approving with nits recorded.

---
**Status: APPROVED** - No blocking findings, 3 nit(s). See inline comments.
*Audited by woostack-review · Host: oh-my-pi · Provider: anthropic · Model: claude-fable-5*



skills/woostack-init/scripts/gen-omp-agents.sh
howarewoo: Fixed in 6f1054a.

skills/woostack-init/scripts/gen-omp-agents.sh
howarewoo: Fixed in 6f1054a.

howarewoo:
### Merge activity

* **Jul 10, 2:52 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/howarewoo/woostack/466).
* **Jul 10, 2:52 AM UTC**: @howarewoo merged this pull request with [Graphite](https://app.graphite.com/github/pr/howarewoo/woostack/466).

PR #495, opened 2026-07-12
Design and plan configurable Linear artifact backend

howarewoo:
## Goal

Define and plan a backward-compatible artifact backend that keeps Markdown as the default while allowing woostack workflows to use Linear as the authoritative spec and plan store through its GraphQL API.

## Summary

- Specifies configurable Markdown and Linear backends, mapping each Linear feature to one project, one spec document, and ordered increment issues.
- Defines a shared direct GraphQL adapter with native lifecycle states, fail-closed errors, idempotent mutations, read-back receipts, and complete command-surface coverage.
- Records the hardened decision to preserve dependency DAGs and fault-isolated overnight tracks while keeping execution sequential.
- Adds a ready implementation plan decomposed into 10 independently reviewable, PR-sized increments.

### Automated

- [ ] `bash skills/woostack-doctor/scripts/checks/doc-type.sh .` (passed with no findings)
- [ ] `bash skills/woostack-doctor/scripts/checks/status-enum.sh .` (passed with no findings)
- [ ] `bash skills/woostack-doctor/scripts/checks/spec-plan-backlink.sh .` (passed with no findings)
- [ ] `bash skills/woostack-doctor/scripts/checks/plan-source.sh .` (completed; reported only an unrelated pre-existing warning for `.woostack/plans/2026-06-14-impeccable-integration.md`)

### Manual

**Before merge**

- [ ] Review the specification and confirm the 10-increment plan covers backend selection, API safety, Linear resources, workflow gates, execution, status reconciliation, doctor/provenance, remaining readers, and authored documentation.
- [ ] Confirm dependency DAGs remain representable while supervised and overnight execution preserve the existing sequential policy.

Spec: .woostack/specs/2026-07-11-linear-artifact-backend.md

PR #497, opened 2026-07-12
Add artifact backend resolver and Markdown adapter

howarewoo:
## Goal

Introduce the artifact-backend resolver and normalized, read-only Markdown adapter while preserving Markdown as the backward-compatible default.

## Summary

- Resolve the configured spec/plan backend and validate complete, secret-free Linear configuration plus canonical repository identity.
- Keep Markdown as the default and add a normalized, read-only Markdown feature model.
- Preserve existing source joins and increment heading forms while ignoring fenced examples, rejecting ambiguous identities, and enforcing artifact containment.
- Add deterministic offline coverage for backend configuration, secret rejection, repository identity, Markdown compatibility, and parser edge cases.

### Automated

- [ ] `bash skills/woostack-init/scripts/tests/test-artifact-backends.sh` (175 passed, 0 failed)
- [ ] `bash skills/woostack-init/scripts/tests/run-tests.sh` (all 12 suites passed)
- [ ] `bash -n` on changed shell scripts (passed)
- [ ] `shellcheck` on changed shell scripts (passed)
- [ ] Task-scoped spec-compliance and code-quality reviews (passed)

### Manual

**Before merge**

- [ ] Confirm the diff contains no credentials or secret Linear configuration values.
- [ ] Confirm the PR is stacked on `feature/linear-artifact-backend`.

Spec: .woostack/specs/2026-07-11-linear-artifact-backend.md

### DIFF
--- a/site/content/docs/configuration.mdx
+++ b/site/content/docs/configuration.mdx
@@ -6,13 +6,16 @@ description: Every key in the .woostack/config.json file, what it tunes, its typ
 woostack reads optional settings from `.woostack/config.json` at the repository root.
 [woostack-init](/docs/skills/woostack-init) creates the file, and every key is optional: when a
 key is absent, the skill that reads it falls back to a built-in default. The init template
-ships four top-level keys: `models`, `review`, `respond`, and `status`.
-[woostack-doctor](/docs/skills/woostack-doctor) checks that all four are present and validates response metadata without querying a provider.
+ships five top-level keys: `artifacts`, `models`, `review`, `respond`, and `status`.
+[woostack-doctor](/docs/skills/woostack-doctor) checks the scaffolded namespaces and validates
+response metadata without querying a provider.
 
-There are eight top-level settings:
+There are ten top-level settings:
 
 | Setting | Read by | Tunes |
 | --- | --- | --- |
+| `artifacts` | spec/plan workflows | selects the Markdown or Linear artifact backend |
+| `linear` | spec/plan workflows when `artifacts.specPlan` is `linear` | secret-free Linear workspace, repository, and lifecycle mappings |
 | `review` | [woostack-review](/docs/skills/woostack-review) | the PR review engine: angles, noise, validation |
 | `models` | [woostack-review](/docs/skills/woostack-review), [woostack-audit](/docs/skills/woostack-audit), and local subagent drivers | shared model selection and tier parameters |
 | `audit` | [woostack-audit](/docs/skills/woostack-audit) | standing-code audit angles, filtering, chunking, and output |
@@ -29,6 +32,30 @@ explained in the sections below.
 
 ```json
 {
+  "artifacts": { "specPlan": "markdown" },
+  "linear": {
+    "workspace": "acme",
+    "team": "ENG",
+    "repository": "acme/widgets",
+    "projectStatuses": {
+      "draft": "Draft",
+      "hardened": "Hardened",
+      "approved": "Approved",
+      "planning": "Planning",
+      "ready": "Ready",
+      "executing": "In Progress",
+      "inReview": "In Review",
+      "done": "Completed",
+      "abandoned": "Canceled"
+    },
+    "issueStates": {
+      "planned": "Backlog",
+      "executing": "In Progress",
+      "inReview": "In Review",
+      "done": "Done",
+      "blocked": "Blocked"
+    }
+  },
   "review": {
     "severity_floor": "high",
     "nits": true,
@@ -67,6 +94,38 @@ explained in the sections below.
   "base_branch": "main"
 }
 ```
+
+## Artifact backend
+
+`artifacts.specPlan` accepts `markdown` (the default) or the reserved `linear` selector. Markdown
+reads tracked `.woostack/specs/*.md` and `.woostack/plans/*.md` files. This increment validates
+the complete, secret-free `linear` namespace shown above; later stacked increments add Linear
+project and issue persistence before the selector becomes operational.
+
+| Key | Required for `linear` | Meaning |
+| --- | --- | --- |
+| `linear.workspace` | yes | Linear workspace slug or identifier. |
+| `linear.team` | yes | Team key or identifier used for increment issues. |
+| `linear.repository` | when no GitHub `origin` can be resolved | Canonical `owner/repository` identity. An explicit value takes precedence over Git remote discovery. |
+| `linear.projectStatuses.draft` | yes | Project status for draft specs. |
+| `linear.projectStatuses.hardened` | yes | Project status for hardened specs. |
+| `linear.projectStatuses.approved` | yes | Project status for approved specs. |
+| `linear.projectStatuses.planning` | yes | Project status while a plan is being written. |
+| `linear.projectStatuses.ready` | yes | Project status for an approved plan awaiting execution. |
+| `linear.projectStatuses.executing` | yes | Project status while increments are being implemented. |
+| `linear.projectStatuses.inReview` | yes | Project status while increment PRs are under review. |
+| `linear.projectStatuses.done` | yes | Project status after every increment is complete. |
+| `linear.projectStatuses.abandoned` | yes | Project status for abandoned work. |
+| `linear.issueStates.planned` | yes | Issue state for planned increments. |
+| `linear.issueStates.executing` | yes | Issue state for active increments. |
+| `linear.issueStates.inReview` | yes | Issue state for increments under review. |
+| `linear.issueStates.done` | yes | Issue state for completed increments. |
+| `linear.issueStates.blocked` | yes | Issue state for blocked increments. |
+
+Only the keys listed above are accepted; unknown keys fail closed. The namespace must not contain
+credentials at any depth. Keep API keys, tokens, authorization headers, passwords, private or
+access keys, client secrets, credential files, and provider authentication in the host's native
+secret store.
 
 ## Response
 
--- a/skills/woostack-init/SKILL.md
+++ b/skills/woostack-init/SKILL.md
@@ -44,17 +44,24 @@ Two callers:
    | `.woostack/wisdom/.gitkeep` | `templates/wisdom/.gitkeep` |
    | `.woostack/respond/` directory | (create empty — tracked sanitized response reports) |
    | `.woostack/respond/.gitkeep` | `templates/respond/.gitkeep` |
-   | `.woostack/config.json` | `templates/config.json` (`{ "models": {}, "review": {}, "respond": {}, "status": { "staleDays": 14 } }`) |
+   | `.woostack/config.json` | `templates/config.json` (`{ "artifacts": { "specPlan": "markdown" }, "models": {}, "review": {}, "respond": {}, "status": { "staleDays": 14 } }`) |
    | `.woostack/.gitignore` | `templates/gitignore` |
    | `.woostack/worktrees/` directory | (create empty — per-PR git worktrees, gitignored) |
 
-   `config.json` ships as `{ "models": {}, "review": {}, "respond": {}, "status": { "staleDays": 14 } }`.
-   Each tool owns its namespace. The optional `respond` namespace accepts only non-secret workflow
-   defaults: `provider`, `environment`, `window`, `max_groups`, and `remediation`; provider
-   credentials remain in provider-native stores. The `status` namespace holds `staleDays`
-   (default 14 — the age in days past which an executing spec is flagged stale on the
-   `/woostack-status` board), defined in
-   [../woostack-status/references/conventions.md](../woostack-status/references/conventions.md).
+   `config.json` ships as `{ "artifacts": { "specPlan": "markdown" }, "models": {}, "review": {},
+   "respond": {}, "status": { "staleDays": 14 } }`. Each tool owns its namespace.
+   `artifacts.specPlan` selects the spec/plan backend: `markdown` keeps tracked `.woostack`
+   artifacts; the reserved `linear` selector currently validates secret-free configuration only.
+   Later stacked increments add Linear project and issue persistence before workflows consume it.
+   The accepted keys and lifecycle mappings are defined in the
+   [artifact backend configuration reference](references/artifact-backends.md). Provider
+   credentials never belong in this file.
+
+   The optional `respond` namespace accepts only non-secret workflow defaults: `provider`,
+   `environment`, `window`, `max_groups`, and `remediation`; provider credentials remain in
+   provider-native stores. The `status` namespace holds `staleDays` (default 14 — the age in
+   days past which an executing spec is flagged stale on the `/woostack-status` board), defined
+   in [../woostack-status/references/conventions.md](../woostack-status/references/conventions.md).
 
    The optional top-level `base_branch` key sets the integration/trunk branch that base branches
    are cut from and PRs target; unset, it auto-detects the remote default (`origin/HEAD`, else
--- a/skills/woostack-init/scripts/artifacts/markdown.sh
+++ b/skills/woostack-init/scripts/artifacts/markdown.sh
@@ -35,49 +35,141 @@ read_field() {
 
 # woostack-defer(increment 5): workflow skills begin consuming the backend adapter in increment 5
 feature() {
-  local requested="$1" spec_dir spec_path woo_root basename stem plan_dir plan_path
-  spec_dir="$(cd "$(dirname "$requested")" 2>/dev/null && pwd -P)" || fail "spec directory not found"
+  local requested="$1" requested_dir repo_hint repo_root spec_dir spec_path woo_root
+  local basename stem plan_dir plan_path spec_slug spec_name
+  requested_dir="$(dirname "$requested")"
+  [[ "$(basename "$requested_dir")" == 'specs' && "$(basename "$(dirname "$requested_dir")")" == '.woostack' ]] \
+    || fail "spec path must be under .woostack/specs"
+  repo_hint="$(dirname "$(dirname "$requested_dir")")"
+  [[ ! -L "$repo_hint/.woostack" && ! -L "$repo_hint/.woostack/specs" ]] \
+    || fail "spec directory symlinks are not allowed"
+  repo_root="$(git -C "$repo_hint" rev-parse --show-toplevel 2>/dev/null)" \
+    || fail "repository root not found"
+  repo_root="$(cd "$repo_root" && pwd -P)"
+  spec_dir="$(cd "$requested_dir" 2>/dev/null && pwd -P)" || fail "spec directory not found"
+  [[ "$spec_dir" == "$repo_root/.woostack/specs" ]] || fail "spec path must be under repository .woostack/specs"
   spec_path="$spec_dir/$(basename "$requested")"
   [[ ! -L "$requested" && ! -L "$spec_path" ]] || fail "spec symlinks are not allowed"
   validate_frontmatter "$spec_path"
   [[ "$(read_field "$spec_path" type)" == 'spec' ]] || fail "artifact is not a spec"
 
-  [[ "$(basename "$spec_dir")" == 'specs' && "$(basename "$(dirname "$spec_dir")")" == '.woostack' ]] \
-    || fail "spec path must be under .woostack/specs"
-  woo_root="$(dirname "$spec_dir")"
+  woo_root="$repo_root/.woostack"
   basename="$(basename "$spec_path")"
   stem="${basename%.md}"
+  spec_slug="${stem#????-??-??-}"
+  spec_name="$(read_field "$spec_path" name)"
   plan_dir="$woo_root/plans"
   [[ -d "$plan_dir" ]] || fail "matching plan not found"
   [[ ! -L "$plan_dir" ]] || fail "plan symlink escapes are not allowed"
 
   local -a matches=()
-  local candidate line canonical legacy
+  local candidate candidate_base candidate_slug candidate_source canonical canonical_md legacy
   canonical="**Source:** [[specs/$stem]]"
+  canonical_md="**Source:** [[specs/$basename]]"
   legacy="**Source:** .woostack/specs/$basename"
   shopt -s nullglob
   for candidate in "$plan_dir"/*.md; do
     [[ ! -L "$candidate" ]] || fail "plan symlinks are not allowed"
-    line="$(awk '
-      $0 == "---" { fences++; next }
-      fences >= 2 && $0 !~ /^[[:space:]]*$/ {
-        value=$0
+    if awk -v canonical="$canonical" -v canonical_md="$canonical_md" -v legacy="$legacy" '
+      function trim(value) {
         sub(/^[[:space:]]*/, "", value)
         sub(/[[:space:]]*$/, "", value)
-        print value
-        exit
+        return value
       }
-    ' "$candidate")"
-    if [[ "$line" == "$canonical" || "$line" == "$legacy" ]]; then
+      function matches_token(value, token) {
+        return index(value, token) == 1 &&
+          (length(value) == length(token) || substr(value, length(token) + 1, 1) ~ /[[:space:]]/)
+      }
+      function run_length(value, marker, count) {
+        marker=substr(value, 1, 1)
+        if (marker != "`" && marker != "~") return 0
+        count=1
+        while (substr(value, count + 1, 1) == marker) count++
+        return count
+      }
+      $0 == "---" && frontmatter_fences < 2 { frontmatter_fences++; next }
+      frontmatter_fences < 2 { next }
+      {
+        value=$0
+        sub(/^[[:space:]]*/, "", value)
+        run=run_length(value)
+        marker=substr(value, 1, 1)
+        rest=substr(value, run + 1)
+        if (in_fence) {
+          if (marker == fence_marker && run >= fence_length && rest ~ /^[[:space:]]*$/) {
+            in_fence=0
+          }
+          next
+        }
+        if (run >= 3) {
+          in_fence=1
+          fence_marker=marker
+          fence_length=run
+          next
+        }
+        value=trim($0)
+        if (matches_token(value, canonical) ||
+            matches_token(value, canonical_md) ||
+            matches_token(value, legacy)) found=1
+      }
+      END { exit !found }
+    ' "$candidate"; then
       matches+=("$candidate")
     fi
   done
-  shopt -u nullglob
 
   if (( ${#matches[@]} == 0 )); then
-    candidate="$plan_dir/$basename"
-    [[ -f "$candidate" ]] && matches+=("$candidate")
+    for candidate in "$plan_dir"/*.md; do
+      candidate_base="$(basename "$candidate" .md)"
+      candidate_slug="${candidate_base#????-??-??-}"
+      candidate_source="$(read_field "$candidate" source)"
+      if [[ -n "$candidate_source" && "$candidate_source" != ".woostack/specs/$basename" ]]; then
+        continue
+      fi
+      if awk '
+        function trim(value) {
+          sub(/^[[:space:]]*/, "", value)
+          sub(/[[:space:]]*$/, "", value)
+          return value
+        }
+        function run_length(value, marker, count) {
+          marker=substr(value, 1, 1)
+          if (marker != "`" && marker != "~") return 0
+          count=1
+          while (substr(value, count + 1, 1) == marker) count++
+          return count
+        }
+        $0 == "---" && frontmatter_fences < 2 { frontmatter_fences++; next }
+        frontmatter_fences < 2 { next }
+        {
+          value=$0
+          sub(/^[[:space:]]*/, "", value)
+          run=run_length(value)
+          marker=substr(value, 1, 1)
+          rest=substr(value, run + 1)
+          if (in_fence) {
+            if (marker == fence_marker && run >= fence_length && rest ~ /^[[:space:]]*$/) in_fence=0
+            next
+          }
+          if (run >= 3) {
+            in_fence=1
+            fence_marker=marker
+            fence_length=run
+            next
+          }
+          if (trim($0) ~ /^\*\*Source:\*\*/) found=1
+        }
+        END { exit !found }
+      ' "$candidate"; then
+        continue
+      fi
+      if [[ "$candidate_slug" == "$spec_slug" || ( -n "$spec_name" && "$candidate_slug" == "$spec_name" ) ]]; then
+        matches+=("$candidate")
+      fi
+    done
   fi
+  shopt -u nullglob
+
   (( ${#matches[@]} > 0 )) || fail "matching plan not found"
   (( ${#matches[@]} == 1 )) || fail "multiple plans join to spec"
   plan_path="${matches[0]}"
@@ -106,23 +198,67 @@ with open(plan_path, encoding="utf-8") as handle:
 
 heading = re.compile(r"^## Increment ([0-9]+)(?::| [-—] | \([^)\r\n]+\):)")
 checkbox = re.compile(r"^[ \t]*-[ \t]+\[([ xX])\]")
+
+frontmatter_fences = 0
+body_start = None
+for index, line in enumerate(lines):
+    if line.rstrip("\r\n") == "---":
+        frontmatter_fences += 1
+        if frontmatter_fences == 2:
+            body_start = index + 1
+            break
+body_lines = lines[body_start:]
+
+def fence_marker(line):
+    stripped = line.lstrip()
+    if not stripped or stripped[0] not in (chr(96), "~"):
+        return None
+    marker = stripped[0]
+    run = len(stripped) - len(stripped.lstrip(marker))
+    if run < 3:
+        return None
+    return marker, run, stripped[run:].rstrip("\r\n")
+
+def checkbox_values(section_lines):
+    values = []
+    active_marker = None
+    active_length = 0
+    for line in section_lines:
+        fence = fence_marker(line)
+        if fence is not None:
+            marker, run, remainder = fence
+            if active_marker is None:
+                active_marker = marker
+                active_length = run
+            elif marker == active_marker and run >= active_length and not remainder.strip():
+                active_marker = None
+                active_length = 0
+            continue
+        if active_marker is None and (match := checkbox.match(line)):
+            values.append(match.group(1))
+    return values
+
 sections = []
 current = None
-in_fence = False
-fence_length = 0
+active_marker = None
+active_length = 0
 
-for line in lines:
-    stripped = line.lstrip()
-    tick_run = len(stripped) - len(stripped.lstrip(chr(96)))
-    if tick_run >= 3:
-        if not in_fence:
-            in_fence = True
-            fence_length = tick_run
-        elif tick_run >= fence_length:
-            in_fence = False
-            fence_length = 0
+for line in body_lines:
+    fence = fence_marker(line)
+    if fence is not None:
+        marker, run, remainder = fence
+        if active_marker is None:
+            active_marker = marker
+            active_length = run
+        elif marker == active_marker and run >= active_length and not remainder.strip():
+            active_marker = None
+            active_length = 0
+        if current is not None:
+            current["lines"].append(line)
         continue
-    if in_fence:
+    if active_marker is not None:
+        if current is not None:
+            current["lines"].append(line)
         continue
     match = heading.match(line.rstrip("\r\n"))
     if match:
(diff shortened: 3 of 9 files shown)


howarewoo:
* **#506** <a href="https://app.graphite.com/github/pr/howarewoo/woostack/506?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#505** <a href="https://app.graphite.com/github/pr/howarewoo/woostack/505?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#504** <a href="https://app.graphite.com/github/pr/howarewoo/woostack/504?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#503** <a href="https://app.graphite.com/github/pr/howarewoo/woostack/503?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#502** <a href="https://app.graphite.com/github/pr/howarewoo/woostack/502?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#501** <a href="https://app.graphite.com/github/pr/howarewoo/woostack/501?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#500** <a href="https://app.graphite.com/github/pr/howarewoo/woostack/500?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#499** <a href="https://app.graphite.com/github/pr/howarewoo/woostack/499?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#498** <a href="https://app.graphite.com/github/pr/howarewoo/woostack/498?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#497** <a href="https://app.graphite.com/github/pr/howarewoo/woostack/497?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/howarewoo/woostack/497?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#495** <a href="https://app.graphite.com/github/pr/howarewoo/woostack/495?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

skills/woostack-init/scripts/artifacts/markdown.sh
```diff
@@ -0,0 +1,201 @@
...
+  local -a matches=()
+  local candidate line canonical legacy
+  canonical="**Source:** [[specs/$stem]]"
+  legacy="**Source:** .woostack/specs/$basename"
+  shopt -s nullglob
+  for candidate in "$plan_dir"/*.md; do
+    [[ ! -L "$candidate" ]] || fail "plan symlinks are not allowed"
+    line="$(awk '
+      $0 == "---" { fences++; next }
+      fences >= 2 && $0 !~ /^[[:space:]]*$/ {
```
howarewoo: **Nit: Legacy plan joins are no longer preserved**

The join records only the first non-blank line after frontmatter and, if that is not a source link, falls back only to an identical full filename. Existing valid plans may put their canonical `**Source:**` link after a title or rely on the existing slug/name fallback. For example, `.woostack/plans/2026-06-02-review-metrics-overlap.md` links to the `2026-06-02-review-metrics-tokens-overlap` spec on line 18, but this adapter reports `matching plan not found` for that spec. The new default Markdown adapter therefore rejects a join that the current reader resolves.

Fix: Search each plan body for an exact canonical or legacy `**Source:**` line outside fenced examples, and preserve the existing date-stripped slug/spec-name fallback when no source line matches.

<sub>— <strong>MEDIUM · NIT</strong> · flagged by the <code>bugs</code> agent</sub>

skills/woostack-init/scripts/artifacts/markdown.sh
```diff
@@ -0,0 +1,201 @@
...
+heading = re.compile(r"^## Increment ([0-9]+)(?::| [-—] | \([^)\r\n]+\):)")
+checkbox = re.compile(r"^[ \t]*-[ \t]+\[([ xX])\]")
+sections = []
+current = None
+in_fence = False
+fence_length = 0
+
+for line in lines:
+    stripped = line.lstrip()
+    tick_run = len(stripped) - len(stripped.lstrip(chr(96)))
```
howarewoo: **Nit: Tilde-fenced examples become real increments**

Fence detection counts only backticks. A valid Markdown `~~~markdown` example containing `## Increment 99: Example` is therefore parsed as an actual increment and its example checkboxes affect normalized progress; a direct probe returns ordinals 1 and 99 instead of only 1. This makes the normalized model depend on which valid Markdown fence delimiter an author chose.

Fix: Track both backtick and tilde fence delimiters, including the opening delimiter type and run length, and ignore headings and checkboxes until a matching valid closing fence is reached.

<sub>— <strong>MEDIUM · NIT</strong> · flagged by the <code>bugs</code> agent</sub>

skills/woostack-init/scripts/artifacts/markdown.sh
```diff
@@ -0,0 +1,201 @@
...
+            sections.append(current)
+        current = {"ordinal": int(match.group(1)), "lines": []}
+    elif current is not None:
+        current["lines"].append(line)
+if current is not None:
+    sections.append(current)
+
+ordinals = [section["ordinal"] for section in sections]
+if not ordinals:
+    raise SystemExit("plan has no increments")
```
howarewoo: **Nit: Headingless legacy plans cannot be normalized**

The adapter unconditionally rejects any plan without a `## Increment N` heading, although the existing Markdown reader accepts older single-plan documents and derives their progress from checkboxes. Exercising the adapter against the repository's current spec/plan pairs fails 16 existing pairs, including `2026-06-02-woostack-init.md`, with `plan has no increments`. This breaks backward compatibility for repositories created before increment headings became the convention.

Fix: Normalize a headingless legacy plan as one implicit increment, preserving its body and checkbox-derived status, while continuing to reject duplicate explicit ordinals.

<sub>— <strong>MEDIUM · NIT</strong> · flagged by the <code>bugs</code> agent</sub>

skills/woostack-init/templates/config.json
```diff
@@ -1,4 +1,7 @@
 {
+  "artifacts": {
```
howarewoo, line 2: **Document the artifact backend configuration**

The init template now ships `artifacts.specPlan`, and the new resolver accepts a complete `linear` namespace, but the published documentation still describes the old configuration. `skills/woostack-init/SKILL.md` twice promises the previous four-key skeleton, `skills/woostack-init/references/memory.md` repeats it, and `site/content/docs/configuration.mdx` says init ships four keys while its purported complete example omits both `artifacts` and the Linear mappings. The command and configuration references are therefore out of sync with the scaffold and do not tell users how to select or configure the new backend.

Fix: Update the init skill, its workspace reference, and the authored configuration page in this PR: show `artifacts.specPlan` in the default skeleton, add `artifacts` to the top-level key inventory and complete example, and document or cross-link the required secret-free `linear` workspace, team, repository, project-status, and issue-state mappings.

<sub>— <strong>HIGH · BLOCKING</strong> · flagged by the <code>docs</code> agent</sub>

skills/woostack-init/scripts/artifacts/markdown.sh
```diff
@@ -0,0 +1,201 @@
...
+      print value
+      exit
+    }
+  ' "$file"
+}
+
+# woostack-defer(increment 5): workflow skills begin consuming the backend adapter in increment 5
+feature() {
+  local requested="$1" spec_dir spec_path woo_root basename stem plan_dir plan_path
+  spec_dir="$(cd "$(dirname "$requested")" 2>/dev/null && pwd -P)" || fail "spec directory not found"
```
howarewoo: **Nit: Canonicalization hides directory symlink escapes**

`pwd -P` resolves a symlinked `.woostack/specs` directory before the later basename-only containment check, while `-L` checks only the final file. A crafted checkout can make `.woostack/specs` point to a sibling repository's `.woostack/specs`; invoking the adapter on the apparent local spec then passes these guards and returns the sibling spec body and joined plan data in JSON.

Fix: Anchor the adapter to a trusted canonical repository root and require the resolved spec directory to equal `<root>/.woostack/specs`; reject symlinked `.woostack` or `specs` path components before reading either artifact.

<sub>— <strong>MEDIUM · NIT</strong> · flagged by the <code>security</code> agent</sub>

skills/woostack-init/scripts/tests/test-artifact-backends.sh
```diff
@@ -0,0 +1,420 @@
...
+  if output="$(bash "$SCRIPT" "$repo" 2>&1)"; then
+    fail "resolver should reject invalid config"
+    RUN_FAILURE_OUTPUT="$output"
+    return
+  fi
+  pass
+  RUN_FAILURE_OUTPUT="$output"
+}
+
+repo="$(make_repo default-markdown)"
```
howarewoo, line 74: **Nit: Missing-config compatibility is untested**

The case named `default-markdown` still writes an empty config file before invoking the resolver, so it never exercises the newly added no-file branch. A regression that fails whenever `.woostack/config.json` is absent would pass this suite even though the resolver explicitly treats that state as Markdown-compatible.

Fix: Add a repository case that does not create `.woostack/config.json`, then assert successful exit and the canonical Markdown resolver JSON.

skills/woostack-init/scripts/tests/test-artifact-backends.sh
```diff
@@ -0,0 +1,420 @@
...
+actual="$(bash "$MARKDOWN" feature "$spec_path")"
+assert_eq "$(jq -c 'keys' <<<"$actual")" '["backend","feature","increments","spec"]' "Markdown model has deterministic top-level fields"
+assert_eq "$(jq -c '.backend' <<<"$actual")" '"markdown"' "Markdown model identifies its backend"
+assert_eq "$(jq -c '.feature | {id,url,title,status,branch}' <<<"$actual")" \
+  "$(jq -cn --arg id ".woostack/specs/$basename.md" --arg title 'feature.v2+api_core' --arg branch 'feature/feature.v2+api_core' '{id:$id,url:null,title:$title,status:"executing",branch:$branch}')" \
+  "canonical wikilink preserves feature identity, lifecycle status, and branch"
+assert_eq "$(jq -c '.spec | {id,url,content:(.content | contains("Spec body.")),revisionLength:(.revision | length)}' <<<"$actual")" \
+  "$(jq -cn --arg id ".woostack/specs/$basename.md" '{id:$id,url:null,content:true,revisionLength:64}')" \
+  "spec content and deterministic revision are normalized"
+assert_eq "$(jq -c '[.increments[] | {ordinal,status,dependencies,branch,pullRequest}]' <<<"$actual")" \
```
howarewoo: **Nit: Increment contract fields escape assertion**

The increment projection checks only `ordinal`, `status`, `dependencies`, `branch`, and `pullRequest`, discarding the normalized model's `id`, `identifier`, and `content`. The suite therefore remains green if the adapter omits those fields, generates unstable IDs, or returns empty section content, all of which break consumers of the shared feature model.

Fix: Assert the complete increment objects, including deterministic Markdown IDs, null identifiers, and representative content boundaries for each parsed section.

skills/woostack-init/scripts/tests/test-artifact-backends.sh
```diff
@@ -0,0 +1,420 @@
...
+  '[{"ordinal":1,"status":"done","dependencies":[],"branch":null,"pullRequest":null},{"ordinal":2,"status":"executing","dependencies":[],"branch":null,"pullRequest":null},{"ordinal":3,"status":"planned","dependencies":[],"branch":null,"pullRequest":null},{"ordinal":4,"status":"planned","dependencies":[],"branch":null,"pullRequest":null}]' \
+  "out-of-order plan sections are sorted by ordinal with matching checkbox-derived statuses"
+assert_eq "$(shasum -a 256 "$spec_path" | cut -d ' ' -f 1)" "$spec_before" "adapter does not rewrite the spec"
+assert_eq "$(shasum -a 256 "$repo/.woostack/plans/$basename.md" | cut -d ' ' -f 1)" "$plan_before" "adapter does not rewrite the plan"
+
+repo="$(make_repo markdown-legacy)"
```
howarewoo, line 367: **Nit: Basename plan fallback has no coverage**

The adapter adds an observable fallback that joins `.woostack/plans/<spec-basename>.md` when no canonical or legacy `Source` line matches, but the suite covers only explicit source lines, missing plans, and duplicate source matches. That compatibility path can regress without any test failure.

Fix: Add a spec and same-basename plan with no `Source` line, assert that normalization succeeds, and retain the existing missing-plan case to prove the fallback stays narrowly scoped.

howarewoo (commented):

The full six-angle swarm and both adversarial validators kept eight findings: one blocking documentation-sync defect and seven non-blocking nits covering Markdown compatibility, containment, and test gaps.

---
**Status: CHANGES REQUESTED** — 1 blocking + 0 non-blocking finding(s) (1 HIGH, 0 MEDIUM, 0 LOW) + 7 nit(s). See inline comments.
*Audited by woostack-review · Host: omp · Provider: openai · Model: openai-codex/gpt-5.6-sol*



skills/woostack-init/scripts/artifacts/markdown.sh
howarewoo: Fixed in a11c687. Plan lookup now finds exact canonical or legacy Source lines anywhere outside fenced examples, then preserves date-stripped slug and spec-name fallback joins.

skills/woostack-init/scripts/artifacts/markdown.sh
howarewoo: Fixed in a11c687. Increment parsing now tracks backtick and tilde fence markers and run lengths, ignores fenced headings and checkboxes, and preserves fenced content.

skills/woostack-init/scripts/artifacts/markdown.sh
howarewoo: Fixed in a11c687. Headingless legacy plans now normalize as one implicit increment with body preservation and checkbox-derived status.

skills/woostack-init/templates/config.json
howarewoo: Fixed in a11c687. The init skill, workspace reference, and authored configuration page now include artifacts.specPlan and the complete secret-free Linear workspace, team, repository, project-status, and issue-state schema.

skills/woostack-init/scripts/artifacts/markdown.sh
howarewoo: Fixed in a11c687. The adapter now anchors artifacts to the canonical Git repository root and rejects symlinked .woostack, specs, spec, plan-directory, and plan paths before reading.

skills/woostack-init/scripts/tests/test-artifact-backends.sh
howarewoo: Fixed in a11c687. The focused backend suite now exercises a repository with no .woostack/config.json and asserts canonical Markdown resolution.

skills/woostack-init/scripts/tests/test-artifact-backends.sh
howarewoo: Fixed in a11c687. Tests now assert every increment field, deterministic IDs, null identifiers, statuses, and representative content boundaries.

skills/woostack-init/scripts/tests/test-artifact-backends.sh
howarewoo: Fixed in a11c687. Tests now cover same-basename and date-stripped slug fallback joins while retaining the missing-plan failure case.

skills/woostack-init/scripts/artifacts/markdown.sh
```diff
@@ -0,0 +1,290 @@
...
+    ' "$candidate"; then
+      matches+=("$candidate")
+    fi
+  done
+
+  if (( ${#matches[@]} == 0 )); then
+    for candidate in "$plan_dir"/*.md; do
+      candidate_base="$(basename "$candidate" .md)"
+      candidate_slug="${candidate_base#????-??-??-}"
+      if [[ "$candidate_slug" == "$spec_slug" || ( -n "$spec_name" && "$candidate_slug" == "$spec_name" ) ]]; then
```
howarewoo, line 166: **Nit: Fallback can join a plan owned by another spec**

When no plan body names the requested spec, the fallback accepts every filename with the same date-stripped slug or spec name without checking whether that candidate explicitly names a different spec. For example, resolving `2026-07-12-foo.md` selects `2026-07-13-foo.md` even when that plan contains `**Source:** [[specs/2026-07-13-foo]]`, so the adapter returns another feature's status and increments instead of reporting that the requested plan is missing.

Fix: Before adding a basename/slug fallback candidate, detect any explicit Source association and exclude the candidate when it targets a different spec; use fallback only for genuinely unassociated legacy plans.

<sub>— <strong>MEDIUM · NIT</strong> · flagged by the <code>bugs</code> agent</sub>

skills/woostack-init/scripts/tests/test-artifact-backends.sh
```diff
@@ -0,0 +1,477 @@
...
+  config="$(jq -c --arg key "$secret_key" --arg value "$secret_value" '.linear.nested = [{safe: true}, {($key): $value}]' <<<"$config")"
+  write_config "$repo" "$config"
+  run_failure "$repo"
+  assert_contains "$RUN_FAILURE_OUTPUT" "$secret_key" "recursive credential-key rejection names only its path"
+  assert_not_contains "$RUN_FAILURE_OUTPUT" "$secret_value" "credential-key diagnostic does not leak its value"
+  secret_index=$((secret_index + 1))
+done
+
+assert_eq "$(jq -c '.artifacts' "$TEMPLATE")" '{"specPlan":"markdown"}' "config template selects Markdown by default"
+assert_eq "$(jq -r '[paths(scalars) as $p | ($p[-1] | tostring) | select(test("apiKey|token|credentialFile"; "i"))] | length' "$TEMPLATE")" '0' "config template contains no credential placeholders"
```
howarewoo: **Nit: Template secret guard omits rejected key classes**

The template guard claims to reject credential placeholders but searches only for `apiKey`, `token`, and `credentialFile`. A template regression adding `password`, `authorization`, `clientSecret`, `credentialsPath`, or another key class rejected by the resolver would still leave this assertion green.

Fix: Exercise the template through `resolve-backend.sh` or derive the template-key check from the same normalized credential-key pattern used by the resolver so every prohibited key class is covered.

skills/woostack-init/scripts/tests/test-artifact-backends.sh
```diff
@@ -0,0 +1,477 @@
...
+spec_path="$repo/.woostack/specs/$basename.md"
+spec_before="$(shasum -a 256 "$spec_path" | cut -d ' ' -f 1)"
+plan_before="$(shasum -a 256 "$repo/.woostack/plans/$basename.md" | cut -d ' ' -f 1)"
+actual="$(bash "$MARKDOWN" feature "$spec_path")"
+assert_eq "$(jq -c 'keys' <<<"$actual")" '["backend","feature","increments","spec"]' "Markdown model has deterministic top-level fields"
+assert_eq "$(jq -c '.backend' <<<"$actual")" '"markdown"' "Markdown model identifies its backend"
+assert_eq "$(jq -c '.feature | {id,url,title,status,branch}' <<<"$actual")" \
+  "$(jq -cn --arg id ".woostack/specs/$basename.md" --arg title 'feature.v2+api_core' --arg branch 'feature/feature.v2+api_core' '{id:$id,url:null,title:$title,status:"executing",branch:$branch}')" \
+  "canonical wikilink preserves feature identity, lifecycle status, and branch"
+assert_eq "$(jq -c '.spec | {id,url,content:(.content | contains("Spec body.")),revisionLength:(.revision | length)}' <<<"$actual")" \
```
howarewoo: **Nit: Revision assertion checks only its length**

The assertion reduces `.spec.revision` to a length check, so a constant or randomly generated 64-character value would pass even though the normalized adapter contract requires a deterministic revision tied to the spec content. The test's current claim of determinism is therefore not protected against regression.

Fix: Assert that `.spec.revision` equals the SHA-256 already captured in `spec_before`, then change the spec content and verify a second adapter invocation produces the corresponding new digest.

howarewoo (commented):

The full review found three non-blocking test and compatibility nits; no blocking findings remain.

---
**Status: APPROVED** — No blocking findings, 3 nit(s). See inline comments.
*Audited by woostack-review · Host: omp · Provider: openai-codex · Model: gpt-5.6-sol*



skills/woostack-init/scripts/artifacts/markdown.sh
howarewoo: Fixed in 27b247e. Date-stripped fallback now excludes plans explicitly associated with another spec, while preserving genuinely unassociated legacy joins.

skills/woostack-init/scripts/tests/test-artifact-backends.sh
howarewoo: Fixed in 27b247e. The template assertion now normalizes keys and applies the resolver's complete credential-key pattern.

skills/woostack-init/scripts/tests/test-artifact-backends.sh
howarewoo: Fixed in 27b247e. The test now compares the revision to the spec SHA-256 digest and verifies that changing content changes the revision.

skills/woostack-init/scripts/artifacts/markdown.sh
```diff
@@ -0,0 +1,331 @@
...
+          next
+        }
+        if (run >= 3) {
+          in_fence=1
+          fence_marker=marker
+          fence_length=run
+          next
+        }
+        value=trim($0)
+        if (value == canonical || value == legacy) found=1
```
howarewoo: **Nit: Annotated source joins are ignored**

The new join parser accepts only an exact source-line string after trimming. Existing readers accept a valid source target followed by whitespace and trailing annotation, and the doctor explicitly preserves that suffix; therefore a plan such as `**Source:** [[specs/2026-07-12-linked]] (stacked feature)` is invisible here. When that plan has a different basename, normalization fails with `matching plan not found` even though the canonical source join is unambiguous.

Fix: Parse the source target with the same exact-slug boundary rules as the established status/doctor readers, allowing trailing text after a closing bracket, whitespace boundary, or optional `.md`, and add an annotated-source regression case with a differing plan basename.

<sub>— <strong>MEDIUM · NIT</strong> · flagged by the <code>bugs</code> agent</sub>

site/content/docs/configuration.mdx
```diff
@@ -68,6 +95,36 @@ explained in the sections below.
 }
 ```
 
+## Artifact backend
+
+`artifacts.specPlan` accepts `markdown` (the default) or `linear`. Markdown reads tracked
+`.woostack/specs/*.md` and `.woostack/plans/*.md` files. Linear stores the same feature model in
```
howarewoo: **Nit: Linear backend docs precede its implementation**

The new configuration reference says selecting `linear` stores the feature model in Linear projects and issues, but this diff only adds backend resolution plus a Markdown adapter. The workflow entry point is explicitly deferred to increment 5, so a user who follows this configuration now will still get the existing Markdown workflow rather than Linear persistence.

Fix: Until the Linear adapter and workflow integration land, describe `linear` as reserved or not yet available and remove the present-tense claims from this page and the init references; publish the operational setup instructions with the increment that actually consumes the backend.

<sub>— <strong>MEDIUM · NIT</strong> · flagged by the <code>docs</code> agent</sub>

skills/woostack-init/scripts/tests/test-artifact-backends.sh
```diff
@@ -0,0 +1,491 @@
...
+basename='2026-07-12-slug-fallback'
+write_feature_pair "$repo" "$basename" ""
+mv "$repo/.woostack/plans/$basename.md" "$repo/.woostack/plans/2026-07-13-slug-fallback.md"
+actual="$(bash "$MARKDOWN" feature "$repo/.woostack/specs/$basename.md")"
+assert_eq "$(jq -r '.feature.id' <<<"$actual")" ".woostack/specs/$basename.md" "date-stripped slug fallback preserves legacy joins"
+
+repo="$(make_repo markdown-slug-fallback-owned)"
+write_feature_pair "$repo" '2026-07-13-owned-slug' '**Source:** [[specs/2026-07-13-owned-slug]]'
+cp "$repo/.woostack/specs/2026-07-13-owned-slug.md" "$repo/.woostack/specs/2026-07-12-owned-slug.md"
+if owned_actual="$(bash "$MARKDOWN" feature "$repo/.woostack/specs/2026-07-12-owned-slug.md" 2>&1)"; then
```
howarewoo: **Nit: Test permits contradictory missing-plan behavior**

This branch treats both a successful empty model and a `matching plan not found` failure as valid for the same explicitly-owned-by-another-spec fixture. The adapter's established missing-plan contract is failure, so a regression that silently normalizes this case to an empty increment list would still leave the suite green.

Fix: Remove the success branch and use `run_markdown_failure`, then assert the `matching plan not found` diagnostic for this fixture.

howarewoo (commented):

The full re-review found three non-blocking compatibility, documentation, and test-contract nits; no blocking findings remain.

---
**Status: APPROVED** — No blocking findings, 3 nit(s). See inline comments.
*Audited by woostack-review · Host: omp · Provider: openai-codex · Model: gpt-5.6-sol*



skills/woostack-init/scripts/artifacts/markdown.sh
howarewoo: Fixed in 5217770. Canonical and legacy Source targets now accept preserved trailing annotations while retaining exact target boundaries; the suite covers a differently named annotated plan.

site/content/docs/configuration.mdx
howarewoo: Fixed in 5217770. PR #497 now describes `linear` as reserved until later stacked increments make persistence operational; the final docs increment restores the operational contract after integration lands.

skills/woostack-init/scripts/tests/test-artifact-backends.sh
howarewoo: Fixed in 5217770. PR #497 now asserts the fail-closed missing-plan contract. Branch 9 updates the same regression in c24568b when that increment intentionally introduces normalized spec-only lifecycles.

skills/woostack-init/scripts/artifacts/markdown.sh
```diff
@@ -0,0 +1,333 @@
...
+          next
+        }
+        if (run >= 3) {
+          in_fence=1
+          fence_marker=marker
+          fence_length=run
+          next
+        }
+        value=trim($0)
+        if (index(value, canonical) == 1 ||
```
howarewoo: **Nit: Optional .md wikilinks no longer resolve**

The established Source reader accepts the Obsidian form `**Source:** [[specs/<basename>.md]]` as well as the extensionless form, but this matcher only recognizes the extensionless string and the bare legacy path. For a plan using the valid `.md` wikilink, the first pass finds no match; the fallback then sees a Source line and deliberately skips slug matching, so `feature` fails with `matching plan not found` instead of returning the normalized artifact.

Fix: Accept both `[[specs/$stem]]` and `[[specs/$basename]]` with the same exact-slug boundary semantics, and add a differently named plan fixture using the `.md` wikilink so slug fallback cannot mask the regression.

<sub>— <strong>MEDIUM · NIT</strong> · flagged by the <code>bugs</code> agent</sub>

skills/woostack-init/SKILL.md
```diff
@@ -44,17 +44,24 @@ Two callers:
...
-   Each tool owns its namespace. The optional `respond` namespace accepts only non-secret workflow
-   defaults: `provider`, `environment`, `window`, `max_groups`, and `remediation`; provider
-   credentials remain in provider-native stores. The `status` namespace holds `staleDays`
-   (default 14 — the age in days past which an executing spec is flagged stale on the
-   `/woostack-status` board), defined in
-   [../woostack-status/references/conventions.md](../woostack-status/references/conventions.md).
+   `config.json` ships as `{ "artifacts": { "specPlan": "markdown" }, "models": {}, "review": {},
+   "respond": {}, "status": { "staleDays": 14 } }`. Each tool owns its namespace.
+   `artifacts.specPlan` selects the spec/plan backend: `markdown` keeps tracked `.woostack`
+   artifacts; `linear` uses Linear projects and issues. The Linear backend requires the
```
howarewoo: **Nit: Linear backend is documented as operational**

The init workflow says selecting `linear` uses Linear projects and issues, but this PR only adds configuration validation and the authored configuration page explicitly says persistence arrives in later stacked increments. A user reading the generated skill reference can therefore expect this selector to store artifacts when it cannot yet do so.

Fix: Describe `linear` as a reserved selector that currently validates its configuration but does not yet persist projects or issues.

```suggestion
   artifacts; `linear` currently validates configuration only and does not persist projects or issues.
```

<sub>— <strong>MEDIUM · NIT</strong> · flagged by the <code>docs</code> agent</sub>

skills/woostack-init/scripts/artifacts/resolve-backend.sh
```diff
@@ -0,0 +1,157 @@
...
+    .linear? // {} |
+    paths as $path |
+    ($path[-1] |
+      if type == "string"
+      then ascii_downcase | gsub("[^a-z0-9]"; "")
+      else ""
+      end
+    ) as $key |
+    select(
+      ($key | test("(apikey|token|credentials?(file|path)|authorization|password|secret)"))
```
howarewoo: **Nit: Private-key credentials bypass rejection**

The recursive credential-name filter does not recognize `privateKey` after normalization. Because `.woostack/config.json` is a tracked file, a private credential stored at a path such as `linear.privateKey` passes resolution instead of stopping the workflow and can then be committed and pushed with the repository.

Fix: Expand the normalized credential-key matcher to reject private-key and access-key names, and add regression cases proving those values are rejected without appearing in diagnostics.

<sub>— <strong>MEDIUM · NIT</strong> · flagged by the <code>security</code> agent</sub>

skills/woostack-init/scripts/artifacts/markdown.sh
```diff
@@ -0,0 +1,333 @@
...
+  [[ "$(read_field "$spec_path" type)" == 'spec' ]] || fail "artifact is not a spec"
+
+  woo_root="$repo_root/.woostack"
+  basename="$(basename "$spec_path")"
+  stem="${basename%.md}"
+  spec_slug="${stem#????-??-??-}"
+  spec_name="$(read_field "$spec_path" name)"
+  plan_dir="$woo_root/plans"
+  [[ -d "$plan_dir" ]] || fail "matching plan not found"
+  [[ ! -L "$plan_dir" ]] || fail "plan symlink escapes are not allowed"
```
howarewoo, line 63: **Nit: Plan-directory symlink guard is untested**

The adapter adds a dedicated rejection branch for a symlinked `.woostack/plans` directory, but the new suite covers only a symlinked specs directory and an individual symlinked plan file. Removing or breaking this directory-level containment check would therefore leave all 186 focused assertions green.

Fix: Add a focused case that replaces `.woostack/plans` with a symlink to an external repository's plans directory, invokes the adapter for an in-repository spec, and asserts that the request fails with the symlink diagnostic.

howarewoo (commented):

The third full review found four non-blocking compatibility, documentation, credential-guard, and test-coverage nits; no blocking findings remain.

---
**Status: APPROVED** — No blocking findings, 4 nit(s). See inline comments.
*Audited by woostack-review · Host: omp · Provider: openai-codex · Model: gpt-5.6-sol*



skills/woostack-init/scripts/artifacts/resolve-backend.sh
howarewoo: **Secret-key blocklist permits credential leaks**

The new credential guard rejects only key names matching a finite regex, while the `linear` object accepts arbitrary nested keys. A developer can place a credential under a common unmatched name such as `auth`, `bearer`, or `privateKey`; the resolver exits successfully, so that secret remains in the tracked `.woostack/config.json` despite the new secret-free invariant and can be exposed to every repository reader.

Fix: Validate the `linear` namespace against an explicit allowlist of the supported keys and nested mapping keys, rejecting every unknown key before returning either backend. Keep the credential-name check as an additional diagnostic layer rather than the sole enforcement boundary.

<sub>— <strong>HIGH · BLOCKING</strong> · flagged by the <code>security</code> agent</sub>

skills/woostack-init/SKILL.md
howarewoo: **Nit: Linear is documented as operational too early**

The public init workflow says that selecting `linear` uses Linear projects and issues, but this PR only adds configuration resolution and the Markdown adapter; the authored configuration page explicitly says persistence arrives in later stacked increments before the selector becomes operational. A user following this workflow can select `linear` expecting artifacts to move to Linear even though no workflow consumes the backend yet.

Fix: Describe `linear` here as a reserved selector that validates its configuration but is not operational until the later persistence and workflow-consumer increments land, matching the configuration reference.

<sub>— <strong>MEDIUM · NIT</strong> · flagged by the <code>docs</code> agent</sub>

skills/woostack-init/SKILL.md
```diff
@@ -44,17 +44,24 @@ Two callers:
...
+   secret-free `linear.workspace`, `linear.team`, optional `linear.repository`, and complete
+   `linear.projectStatuses` and `linear.issueStates` mappings documented in the
+   [configuration reference](../../site/content/docs/configuration.mdx#artifact-backend).
```
howarewoo: **Nit: Configuration reference is outside the skill bundle**

The new Linear workflow points outside `skills/woostack-init/` to the docs application for its mapping contract. Agent Skill progressive disclosure expects supporting reference material to ship with the skill; after installation this target is not part of the skill bundle, so the agent cannot load the promised details.

Fix: Put the backend configuration contract in a descriptive file under `skills/woostack-init/references/` and link that bundled reference directly from this line; keep the site page as user-facing documentation without making it the skill's operational dependency.

<sub>— <strong>MEDIUM · NIT</strong> · flagged by the <code>skills</code> agent</sub>

skills/woostack-init/scripts/artifacts/markdown.sh
howarewoo, line 63: **Nit: Plan-directory containment branch lacks coverage**

The containment cases cover a symlinked specs directory and a symlinked plan file, but never a symlink at `.woostack/plans` itself. The dedicated directory guard can therefore regress while the suite remains green, allowing the adapter to discover and read plans from outside the repository.

Fix: Add a fixture whose `.woostack/plans` directory is a symlink to an external repository directory and assert that the adapter rejects it with the symlink diagnostic.

skills/woostack-init/scripts/artifacts/resolve-backend.sh
```diff
@@ -0,0 +1,157 @@
...
+    ) |
+    $path | map(if type == "number" then "[\(.)]" else . end) | join(".")
+  ) // empty
+' <<<"$config")"
+if [ -n "$secret_path" ]; then
```
howarewoo, line 42: **Nit: Credential rejection is only tested in Linear mode**

Every credential-key fixture is built with `specPlan: linear`, so the suite stays green if recursive secret validation is later moved below the Markdown early return. That would let a default or explicitly Markdown configuration retain `linear.apiKey` or another forbidden credential key despite the namespace-wide rejection contract.

Fix: Add a default- or Markdown-mode fixture with a nested credential-shaped key, then assert that resolution fails, names only the config path, and does not expose the value.

howarewoo (commented):

The full review found one blocking security gap in Linear configuration validation and four non-blocking documentation/test suggestions. Existing unresolved threads also remain open.

---
**Status: CHANGES REQUESTED** — 1 blocking + 0 non-blocking finding(s) (1 HIGH, 0 MEDIUM, 0 LOW) + 4 nit(s). See inline comments.
*Audited by woostack-review · Host: omp · Provider: openai-codex · Model: gpt-5.6-sol*



skills/woostack-init/scripts/artifacts/markdown.sh
howarewoo: Fixed in `550d34b`.

skills/woostack-init/SKILL.md
howarewoo: Fixed in `550d34b`.

skills/woostack-init/scripts/artifacts/resolve-backend.sh
howarewoo: Fixed in `550d34b`.

skills/woostack-init/scripts/artifacts/markdown.sh
howarewoo: Fixed in `550d34b`.

skills/woostack-init/scripts/artifacts/resolve-backend.sh
howarewoo: Fixed in `550d34b`.

skills/woostack-init/SKILL.md
howarewoo: Fixed in `550d34b`.

skills/woostack-init/SKILL.md
howarewoo: Fixed in `550d34b`.

skills/woostack-init/scripts/artifacts/markdown.sh
howarewoo: Fixed in `550d34b`.

skills/woostack-init/scripts/artifacts/resolve-backend.sh
howarewoo: Fixed in `550d34b`.

PR #570, opened 2026-07-28
docs(linear): document collaboration and engineer agents

howarewoo:
## Goal

Align the README and authored docs site with Linear-only collaboration and the engineer-agent operating model.

## Summary

- Rewrite getting-started, configuration, workflows, status, utilities, memory, and building-rules framing.
- Add dedicated engineer-agent documentation and keep the 23-command/26-location skill surface exact.
- Publish copyable Hermes setup and prompt guidance plus the OMP host profile.

### Automated

- `bash site/scripts/linear-only-docs.test.mjs`
- `pnpm -C site test`
- `pnpm -C site build`

### Manual

- Read the getting-started and harness pages as a new adopter; verify authority, identity, and handoff steps are unambiguous.

Spec: .woostack/specs/2026-07-26-linear-only-collaboration.md

PR #569, opened 2026-07-28
feat(execution): add engineer-agent contract

howarewoo:
## Goal

Publish a host-neutral decision-making engineer-agent contract with isolated Hermes and OMP profiles.

## Summary

- Define lead, issue-engineer, coder, reviewer, ownership, collision, escalation, and handoff invariants.
- Give each long-running Hermes/OMP unit a distinct Linear identity and least-privilege credentials.
- Add deterministic behavior corpora for allocation, independent review, escalation, dependency, attribution, and terminal reconciliation.

### Automated

- `bash skills/using-woostack/tests/test-engineer-agent-contract.sh`
- Run the package eval suites covering allocation, collision, review, escalation, and handoff.

### Manual

- Verify Hermes and OMP setup examples cannot share an identity or permit coder self-review.

Spec: .woostack/specs/2026-07-26-linear-only-collaboration.md

PR #568, opened 2026-07-28
feat(context): require verified Linear MCP authority

howarewoo:
## Goal

Bind diagnostics, remediation, review, QA, audit, and read-only utilities to verified Linear context.

## Summary

- Route explicit read-only utilities through exact MCP project/issue context with no local development-authority fallback.
- Require remediation commands to verify the managed issue and current owner before tracked repository mutation.
- Remove custom review/CI artifact context while preserving GitHub review-thread transport and sanitized local diagnostic reports.

### Automated

- Run package validators for ask, audit, QA, respond, review, address-comments, build, and bootstrap.
- `bash skills/woostack-address-comments/scripts/tests/test-address-worker-contract.sh`
- Run the audit smoke and focused respond contract suites.

### Manual

- Verify local diagnostic reports remain non-authoritative and cannot authorize remediation.

Spec: .woostack/specs/2026-07-26-linear-only-collaboration.md

PR #567, opened 2026-07-28
feat(execution): move tracking to Linear MCP

howarewoo:
## Goal

Make Linear MCP the execution, overnight, sweep, and status state plane without local development ledgers.

## Summary

- Drive sequential increment claiming, relations, receipts, worktrees, submission, and `inReview` transitions from managed Linear issues.
- Store overnight and review-sweep evidence as typed Linear updates/comments.
- Reconcile terminal status only from verified merge evidence and remove backend-derived local board state.

### Automated

- Run the execute, execute-overnight, sweep, and status behavior suites.
- Run deterministic relation, ancestry, ownership, receipt, and terminal-state fixtures.

### Manual

- Confirm a dependent issue cannot start until native relation and immutable Git-parent readiness both read back.

Spec: .woostack/specs/2026-07-26-linear-only-collaboration.md

PR #571, opened 2026-07-28
refactor(linear): remove local transport

howarewoo:
## Goal

Complete the Linear-only cutover by deleting legacy transport/backend machinery, providing a loss-safe migration owner, and proving the full product contract.

## Summary

- Remove custom Linear GraphQL/API-key adapters, backend selectors, local spec/plan templates, and retired fixtures.
- Route one-way legacy record migration through `woostack-init --migrate-legacy` with authenticated approvals, stable identity, complete read-back receipts, Git byte recovery, and fail-closed deletion boundaries.
- Reconcile skill descriptions, trigger/eval corpora, generated-page contracts, idempotent relation add/remove planning, and authored migration documentation.
- Preserve GitHub GraphQL only for review-thread operations; all development records use official Linear MCP.

### Automated

- `bash skills/woostack-init/scripts/tests/test-legacy-migration-contract.sh`
- `bash skills/woostack-build/tests/test-linear-plan-contract.sh`
- `bash skills/woostack-eval/scripts/tests/test-critical-corpora.sh`
- `bash skills/woostack-init/scripts/tests/test-linear-only-contract.sh`
- `bash skills/woostack-init/scripts/tests/test-host-references.sh`
- `node --test site/scripts/linear-only-docs.test.mjs`
- Full deterministic package/structural verification: all required commands passed.

### Manual

- Live official Linear MCP acceptance passed for WOO-5, project `3d78c276-5046-42a8-b0f0-ac8e5ec0119c`, WOO-6, and WOO-7: create/read/update/comment/assignment/relation/state/exact-marker recovery all read back; every disposable resource was then read back Canceled.
- Confirm archivedAt remained null because the mounted tools expose cancellation, not immediate archival.

Spec: .woostack/specs/2026-07-26-linear-only-collaboration.md


howarewoo:
* **#571** <a href="https://app.graphite.com/github/pr/howarewoo/woostack/571?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/howarewoo/woostack/571?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#570** <a href="https://app.graphite.com/github/pr/howarewoo/woostack/570?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#569** <a href="https://app.graphite.com/github/pr/howarewoo/woostack/569?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#568** <a href="https://app.graphite.com/github/pr/howarewoo/woostack/568?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#567** <a href="https://app.graphite.com/github/pr/howarewoo/woostack/567?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#566** <a href="https://app.graphite.com/github/pr/howarewoo/woostack/566?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#565** <a href="https://app.graphite.com/github/pr/howarewoo/woostack/565?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#564** <a href="https://app.graphite.com/github/pr/howarewoo/woostack/564?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#562** <a href="https://app.graphite.com/github/pr/howarewoo/woostack/562?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

skills/woostack-init/references/legacy-migration.md
```diff
@@ -0,0 +1,89 @@
...
+adapter, token file, local development authority, or title match.
+
+## 1. Freeze recoverable source bytes
+
+1. Enumerate every tracked legacy file in the requested set with `git ls-files`, rejecting symlinks,
+untracked files, path escapes, malformed records, and a set whose relationships cannot be resolved.
+2. For each file record its repository-relative path, current SHA-256, containing commit SHA, and
+`git show <commit>:<path>` SHA-256. Require the recovered bytes to equal the working-tree bytes
+byte-for-byte. A dirty, missing, untracked, or unrecoverable file blocks the whole set.
+3. Classify the complete set as active, historical-completed, ambiguous, partial, incomplete,
```
howarewoo: **[HIGH · blocking] Resume verified partial migrations instead of deadlocking**

The classifier makes every `partial` set preservation-only and prohibits any new remote mutation. If an interrupted migration has already created the project but has not yet created one planned issue, every rerun observes the same partial state and is forbidden from reaching the later “Create or update only missing managed state” step, contrary to stable-UUID resume behavior. Distinguish untrusted incomplete receipts from verified partial boundaries and allow only missing mutations while keeping deletion blocked.

_Evidence:_ legacy-migration.md:20-23 forbids mutations for partial sets; later procedure requires creating missing managed state.

skills/woostack-init/references/legacy-migration.md
```diff
@@ -0,0 +1,89 @@
...
+unknown-outcome, or foreign. Ambiguous/partial/incomplete/unknown/foreign sets are preservation-only:
+perform no new remote mutation and prove every source path retains its original SHA-256 before
+returning.
+
+## 2. Bind stable managed identities
+
+Allocate one UUIDv4 client ID per intended project, issue, update, comment, and relation before its
+first mutation. Before any remote write, persist the IDs, source-byte ledger, and next boundary in
```
howarewoo: **[HIGH · blocking] Define how resumed migrations retrieve the recovery receipt**

The procedure requires the first invocation to persist a durable receipt outside the repository and a resumed invocation to present that exact receipt, but defines no schema, canonical host-owned location, lookup identity, command argument, or returned resume token. After interruption a new invocation cannot locate stable IDs and the next boundary, while missing receipts are rejected. Define and return a portable atomic receipt handle and validated payload.

_Evidence:_ legacy-migration.md:28-31 requires a durable external receipt without a retrieval/input contract.

skills/woostack-init/references/legacy-migration.md
```diff
@@ -0,0 +1,89 @@
...
+`git show <commit>:<path>` SHA-256. Require the recovered bytes to equal the working-tree bytes
+byte-for-byte. A dirty, missing, untracked, or unrecoverable file blocks the whole set.
+3. Classify the complete set as active, historical-completed, ambiguous, partial, incomplete,
+unknown-outcome, or foreign. Ambiguous/partial/incomplete/unknown/foreign sets are preservation-only:
+perform no new remote mutation and prove every source path retains its original SHA-256 before
+returning.
+
+## 2. Bind stable managed identities
+
+Allocate one UUIDv4 client ID per intended project, issue, update, comment, and relation before its
```
howarewoo: **[MEDIUM · non-blocking] Reconcile native relations by their canonical tuple**

This allocates a client UUID for each intended native relation and directs resume logic to search by every recorded client ID. Canonical managed metadata has client-ID envelopes for projects, issues, updates, and comments, while planning reconciles relations by exact native endpoint/type tuples; no relation dispatcher embeds or searches a migration UUID. Recover relations from the exact endpoint/type tuple and native relation set.

_Evidence:_ legacy-migration.md:27-32 assigns client IDs to relations; canonical planning relation reconciliation uses native relation sets.

skills/woostack-eval/scripts/tests/test-critical-corpora.sh
```diff
@@ -253,36 +760,206 @@ function semanticDigest(value) {
...
+      && !Array.isArray(entry)
+      && same(Object.keys(entry).sort(), ['caseId', 'kind', 'repetition', 'variant'])
+      && typeof entry.caseId === 'string'
+      && ['behavior', 'trigger'].includes(entry.kind)
+      && Number.isInteger(entry.repetition)
+      && entry.repetition > 0
+      && ['baseline', 'candidate'].includes(entry.variant));
+}
+
+function assertNoWorkerFixtureOracle(value, location, pointer = '') {
```
howarewoo: **[HIGH · blocking] Enforce fixture-oracle rejection in the production validator**

The new oracle check exists only inside this shell test's inline JavaScript. The actual validate.mjs fixture path checks containment, tracking, type, and existence but never calls equivalent content validation, while prepare.mjs copies every declared fixture into both worker variants. Consequently a normal woostack-eval validation can accept a declared answer-oracle fixture and expose it to the model; only repositories that separately run this hard-coded critical-corpora test are protected. Move the oracle check into validate.mjs and have this negative case invoke the production validator.

_Evidence:_ test-critical-corpora.sh:802-845 defines assertNoWorkerFixtureOracle locally and calls it only from its own semantic digest. validate.mjs:499-522 and 590-606 validate fixture paths without reading content; prepare.mjs:791-807 and 1413-1417 copies declared fixtures into candidate and baseline worker workspaces.

skills/woostack-init/evals/evals.json
```diff
@@ -162,6 +162,169 @@
...
+        {"id":"migration-create-predelete-complete-set","kind":"final-json-path-equals","pointer":"/preDeleteCompleteSetVerified","expected":true,"critical":true},
+        {"id":"migration-create-predelete-git-bytes","kind":"final-json-path-equals","pointer":"/preDeleteGitRecoveryVerified","expected":true,"critical":true},
+        {"id":"migration-create-deletion","kind":"final-json-path-equals","pointer":"/localDeletions","expected":[".woostack/specs/orders.md"],"critical":true},
+        {"id":"migration-create-no-replay","kind":"final-json-path-equals","pointer":"/createReplayAllowed","expected":false,"critical":true},
+        {"id":"migration-create-source-deleted","kind":"path-absent","path":"fixtures/migration/create-active/project/.woostack/specs/orders.md","critical":true},
+        {"id":"migration-create-git-recovery","kind":"file-sha256-equals","file":"fixtures/migration/create-active/git-recovery/.woostack/specs/orders.md","sha256":"sha256:229fb72e68435fe2111693b02ecabacbcbdeff1d2ba06133b890efff0967d3e9","critical":true}
+      ]
+    },
+    {
+      "id": "migrates-active-records-by-stable-client-id",
```
howarewoo, line 202: **[HIGH · blocking] Remove migration verdicts from worker-visible eval inputs**

The added migration cases tell the worker the classification before it applies the packaged skill: this blocked case enumerates the missing proof and explicitly orders preservation, while the create case describes the complete eligible path and its fixture includes a terminalReceipt whose booleans mirror the asserted output. A model can therefore satisfy the blocked and destructive cases by restating the prompt/receipt without validating the raw operation transcript, so regressions in the actual migration rules can still score as correct. Keep only neutral scenario facts and raw receipts in worker-visible inputs; leave the verdict and derived verification flags exclusively in expected/assertions.

_Evidence:_ evals.json:197 says the supplied rows omit the required proof and commands the worker to preserve every byte; lines 168-193 similarly describe the eligible create path and exact output shape. create-active.json:441-456 supplies complete/currentPhase/currentHeadCount/issueGraphComplete/provenanceVerified/preDeleteCompleteSetVerified/preDeleteGitRecoveryVerified values that directly match the case assertions.

skills/woostack-init/references/legacy-migration.md
```diff
@@ -0,0 +1,89 @@
+# One-way legacy development-record migration
+
+`/woostack-init --migrate-legacy` is the sole routed owner of migration from tracked
+`.woostack/specs/`, `.woostack/plans/`, and `.woostack/fixes/` records to the managed Linear
```
howarewoo: **[HIGH · blocking] Include legacy overnight records in migration**

The migration owner enumerates only `.woostack/specs/`, `.woostack/plans/`, and `.woostack/fixes/`, while doctor emits a blocking `legacy-development-records` finding for any nonempty `.woostack/overnight/` directory. A repository with a tracked legacy overnight handback is therefore told to run `--migrate-legacy`, but this procedure has no path to classify, receipt, or delete that set, so the blocker cannot be cleared without bypassing the loss-safe migration boundary. Include overnight records in classification, Git/provenance verification, and all-or-nothing deletion, and sync the init and public migration guidance.

_Evidence:_ Changed legacy-migration.md lines 3-6 and woostack-init/SKILL.md lines 63-67 name only specs/plans/fixes; site/content/docs/configuration.mdx lines 16-19 repeats that set. skills/woostack-doctor/scripts/checks/config-keys.sh lines 80-87 and woostack-doctor/SKILL.md lines 50-52 explicitly dispatch nonempty overnight directories to the blocking migration finding. The governing spec lines 161-175 includes `.woostack/overnight/` in the one-way migration.

skills/woostack-init/references/legacy-migration.md
```diff
@@ -0,0 +1,89 @@
...
+<commit>:<path>@sha256:<source-sha256> -> linear://project/<client-uuid>
+<commit>:<path>@sha256:<source-sha256> -> linear://issue/<client-uuid>
+```
+
+Historical-completed input additionally requires exact canonical PR attribution, independently
+verified merge commit and ancestry, and a managed completed/accepted receipt. Active input requires
+the complete managed specification and issue graph, current ownership, dependency relations, and
+readable provenance. A boolean summary or mutation response is never proof.
+
+Re-read the complete related resource set by both stable client IDs and native IDs. The terminal
```
howarewoo: **[HIGH · blocking] Repair knowledge provenance before deletion**

The terminal receipt checks only managed resources, the source-to-managed ledger, and Git recovery, then authorizes deletion without enumerating or rewriting preserved memory/wisdom provenance. Existing `source:` values and body links that name a migrated spec, plan, fix, or overnight report will therefore point at deleted files even though the command returns complete; active sources must become stable Linear URIs and historical sources immutable Git/PR provenance, with every preserved link verified before any deletion.

_Evidence:_ Changed legacy-migration.md lines 67-78 defines the terminal receipt and deletion gate without a memory/wisdom read or repair. The governing spec lines 167-175 and implementation plan lines 180-204 require memory/wisdom provenance rewriting and zero deletion on unresolved provenance. memory.md lines 90 and 109 identify historical Markdown sources as migration input, and current tracked `.woostack/memory/` and `.woostack/wisdom/` contain such plan/fix source ledgers.

skills/woostack-init/references/legacy-migration.md
```diff
@@ -0,0 +1,89 @@
...
+Create or update only missing managed state. Preserve stable identities on every retry. After each
+mutation, direct-read the native resource and independently re-list its owner, project membership,
+and relevant relations. Build a source-to-managed ledger entry for every file:
+
+```text
+<commit>:<path>@sha256:<source-sha256> -> linear://project/<client-uuid>
+<commit>:<path>@sha256:<source-sha256> -> linear://issue/<client-uuid>
+```
+
+Historical-completed input additionally requires exact canonical PR attribution, independently
```
howarewoo: **[HIGH · blocking] Keep completed history out of Linear**

There is no Git-only completion path for a historical record that has verified merge and recovery evidence but no managed Linear resource: the procedure says to create missing managed state, requires a managed completed/accepted receipt for `historical-completed`, and allows deletion only after the managed-resource terminal receipt. Such completed history must therefore either be imported or remain blocked, contrary to the active-only cutover. Special-case historical sets so they perform zero Linear mutation, retain immutable Git blob/PR provenance, and can participate in the all-or-nothing deletion gate from Git/GitHub evidence alone.

_Evidence:_ Changed legacy-migration.md lines 53-63 applies managed-state reconciliation and a managed receipt to historical-completed input; lines 74-76 make that terminal receipt a deletion prerequisite. The governing spec lines 48-50 and 161-175 says completed historical records are not imported and Git remains their archive; AC5 lines 259-262 repeats that interrupted reruns never import completed history. README.md lines 88-90 and site/content/docs/configuration.mdx lines 16-19 also currently imply every legacy source waits for managed Linear ownership/receipt.

skills/woostack-ideate/SKILL.md
```diff
@@ -30,15 +30,14 @@ approval before moving on.
...
+- **Write nothing.** The approved design lives in the conversation.
 - **Chain nothing.** Do not invoke `woostack-plan`, `woostack-execute`, or any implementation
   skill yourself.
 - **Hand back.** State that the design is approved and name the next step:
-  - Inside `woostack-build`: return control to its backend split. The caller captures the
-    design as a Markdown spec or Linear managed project/spec according to the selected artifact backend.
-  - Standalone: tell the user the design is ready for `woostack-build` to capture through the
-    configured backend, and stop.
+  - Inside `woostack-build`: return control to the caller, which captures the design as the
+    specification in the verified managed feature project through official host-exposed Linear MCP.
```
howarewoo: **[MEDIUM · blocking] Remove remaining backend-resolution instructions**

The new handback correctly sends the approved design to the one verified Linear feature project, but the same skill still says the caller persists through a selected artifact backend and, in its hard constraints, owns backend resolution. Since this patch removes backend selection and the resolver, an agent receives mutually exclusive terminal instructions and can stop to resolve or invent a backend that no longer exists. Replace both surviving statements with the official-MCP Linear handback used here.

_Evidence:_ Changed lines 37-40 define the Linear-only handback. Unchanged active lines 11-13 still say `selected artifact backend`, and lines 128-129 still make the caller own `backend resolution and persistence`; the canonical authority explicitly says no selectable backend exists.

skills/woostack-visualize/SKILL.md
```diff
@@ -94,17 +89,15 @@ Linear read. It must not silently acquire development context.
 6. **High-stakes self-review.** For architecture, backend, data model, migration, security,
    multi-file, or public-contract renders, verify every claim against its pinned source, offline
    rendering, audience fit, safe encoding, and explicit coverage gaps. Fix or report any failure.
-7. **Write and report.** Write to `.woostack/visuals/YYYY-MM-DD-<slug>-<audience>.html`, or an
-   explicit user path outside all development-record directories. If `.woostack/` is absent, write
-   next to the immutable source or to the explicit path. Report the path and offer to open it; never
-   open a browser without consent.
+7. **Write and report.** Write to `.woostack/visuals/YYYY-MM-DD-<slug>-<audience>.html` or an
+   explicit user path. If `.woostack/` is absent, write next to the immutable source or to the
```
howarewoo: **[MEDIUM · blocking] Keep visualization output out of legacy record directories**

This edit removes the requirement that an explicit output path stay outside development-record directories, so a requested path under `.woostack/specs/`, `.woostack/plans/`, `.woostack/fixes/`, or `.woostack/overnight/` is now allowed. Doctor treats any entry in those directories, regardless of extension or authority disclaimer, as a blocking legacy migration set; one disposable visualization can therefore make subsequent woostack development fail closed. Restore the directory exclusion for explicit output paths.

_Evidence:_ Changed visualize line 93 permits any explicit user path; the deleted text required it to be outside all development-record directories. skills/woostack-doctor/scripts/checks/config-keys.sh lines 80-87 emits `legacy-development-records` for any nonempty one of the four directories without filtering file type, while visualize's default `.woostack/visuals/` path is safe.

howarewoo (commented):
Hephaestus reviewed the exact stacked increment at `7605b612b2f51e7190a17b5f4c67ef84dc8ecb1b`. Result: **BLOCKING FINDINGS** — 10 independently validated findings: 9 blocking, 1 non-blocking. GitHub does not permit an author to request changes on their own PR, so this self-review is submitted as COMMENT while retaining the blocking classification.

The native woostack-review preflight is non-runnable on this stack because the migrated repository config fails artifact-backend resolution. This review used complete local base...head diff coverage plus an independent second adjudication; no native receipt is claimed.

skills/woostack-init/references/legacy-migration.md
howarewoo: Fixed in ef22d94c.

skills/woostack-init/references/legacy-migration.md
howarewoo: Fixed in ef22d94c.

skills/woostack-init/references/legacy-migration.md
howarewoo: Fixed in ef22d94c.

skills/woostack-eval/scripts/tests/test-critical-corpora.sh
howarewoo: Fixed in ef22d94c.

skills/woostack-init/evals/evals.json
howarewoo: Fixed in ef22d94c.

skills/woostack-init/references/legacy-migration.md
howarewoo: Fixed in ef22d94c.

skills/woostack-init/references/legacy-migration.md
howarewoo: Fixed in ef22d94c.

skills/woostack-init/references/legacy-migration.md
howarewoo: Fixed in ef22d94c.

skills/woostack-ideate/SKILL.md
howarewoo: Fixed in ef22d94c.

skills/woostack-visualize/SKILL.md
howarewoo: Fixed in ef22d94c.

skills/woostack-init/references/legacy-migration.md
```diff
@@ -0,0 +1,153 @@
...
+official MCP is the only remote transport; Git is the recovery source. Never use a repository
+adapter, token file, local development authority, or title match.
+
+## 1. Freeze recoverable source bytes
+
+1. Enumerate every tracked legacy file in the requested related set with `git ls-files`, including
+   specifications, plans, fixes, and overnight handbacks. Reject symlinks, untracked files, path
+   escapes, malformed records, and any set whose relationships cannot be resolved.
+2. For each file record its repository-relative path, current SHA-256, containing commit SHA, and
+   `git show <commit>:<path>` SHA-256. Require the recovered bytes to equal the working-tree bytes
```
howarewoo: [High] Make the tracked-file recovery check work on clean Windows checkouts

This compares working-tree bytes directly with the raw `git show <commit>:<path>` blob. With the repository's `text=auto` attributes and a normal Windows checkout, a clean tracked Markdown file has CRLF working-tree bytes while the Git blob has LF bytes, so this preflight blocks migration even though `git status` is clean and Git can restore the file. I reproduced this on the exact stack tip: the tracked spec had zero status lines, `git hash-object --no-filters` = `37472cc...`, blob = `24a3860...`, and `file` reported CRLF. Compare against a Git-filtered temporary checkout (or another normalized Git identity that still proves recoverability) and add a CRLF/Windows fixture; otherwise the one-shot migration is unavailable on a supported host.

skills/woostack-init/references/legacy-migration.md
```diff
@@ -0,0 +1,153 @@
...
+  ledger;
+- no foreign resource or unenumerated legacy path is present; and
+- `git show` still recovers every source byte at its recorded SHA-256.
+
+Local deletion is a separate all-or-nothing final boundary. Immediately before deleting, repeat
+the applicable active managed reads or historical Git/PR reads, relation reconciliation,
+knowledge-link verification, and Git byte recovery. Then delete only the enumerated legacy paths.
+Immediately after deletion, require every path to be absent and recover every deleted file again
+with `git show`, comparing SHA-256 byte-for-byte to the preflight ledger.
+
```
howarewoo: [High] Roll back knowledge provenance when the final deletion boundary aborts

Section 4 has already applied the Linear provenance rewrite before this final source/receipt re-read. If this check is stale or fails, Section 5 says to delete nothing, but it never restores the now-modified knowledge files. That violates this reference's own failure contract and spec AC5, both of which require every local record and the original provenance to remain intact on any evidence/read-back failure. Keep the knowledge rewrite staged until the final boundary succeeds, or restore and hash-verify every original on every post-rewrite failure path, and add a late pre-delete failure fixture.

howarewoo (commented):
Re-review after the addressed threads found two remaining blocking migration issues. GitHub's review principal is the PR author, so this is submitted as COMMENT rather than REQUEST_CHANGES; stack verdict remains NO-GO until both are resolved.

skills/woostack-eval/scripts/validate.mjs
```diff
@@ -44,6 +44,16 @@ const BEHAVIOR_CASE_FIELDS = new Set([
 const TRIGGER_CASE_FIELDS = new Set([
   'id', 'query', 'shouldTrigger', 'expectedSkill', 'conflictsWith',
 ]);
+const WORKER_FIXTURE_ORACLE_KEYS = new Set([
```
howarewoo, line 52: **[High] Reject semantic output oracles, not only a short key blacklist**

The production guard accepts fixtures that expose the answer under unlisted names. For example, `fixtures/migration/ambiguous.json` gives workers both `scenario: "ambiguous"` and `receiptSet.outcome: "ambiguous"`, while the behavior case asserts `classification == "ambiguous"`. The same leakage pattern appears across the new migration fixtures. This makes the critical corpus capable of passing by copying fixture labels instead of deriving the procedure. Remove/neutralize worker-visible classifications and add adversarial validator fixtures for `scenario`, nested `outcome`, `status`, `classification`, and equivalent semantic aliases rather than relying only on this blacklist.

skills/woostack-init/references/legacy-migration.md
```diff
@@ -0,0 +1,153 @@
...
+## 1. Freeze recoverable source bytes
+
+1. Enumerate every tracked legacy file in the requested related set with `git ls-files`, including
+   specifications, plans, fixes, and overnight handbacks. Reject symlinks, untracked files, path
+   escapes, malformed records, and any set whose relationships cannot be resolved.
+2. For each file record its repository-relative path, current SHA-256, containing commit SHA, and
+   `git show <commit>:<path>` SHA-256. Require the recovered bytes to equal the working-tree bytes
+   byte-for-byte. A dirty, missing, untracked, or unrecoverable file blocks the whole set.
+3. Classify the complete set as active, historical-completed, ambiguous, partial, incomplete,
```
howarewoo: **[High] Classify each record before deriving the set action**

The approved spec requires each record to be classified and the migration to present exact active, historical, and ambiguous subsets. This procedure instead assigns one classification to the complete related set and sends a historical set around active reconciliation. A normal related chain can contain an active approved spec plus a historical completed fix/overnight record; neither singular branch can process that safely without importing history or omitting active state. Define per-record classifications, derive a blocked/active/historical set plan from them, and add a mixed active+historical fixture that proves active resources are reconciled, historical records remain Git-only, and deletion stays all-or-nothing.

howarewoo (commented):
Additional follow-up findings from the completed whole-stack review; these are in addition to the two migration blockers already posted. Posting COMMENT because the authenticated reviewer is also the PR author.

skills/woostack-init/references/legacy-migration.md
howarewoo: Fixed in 5a60f1cf853ff1fbc4db8ef82fadf13d02c65848.

skills/woostack-init/references/legacy-migration.md
howarewoo: Fixed in 5a60f1cf853ff1fbc4db8ef82fadf13d02c65848.

skills/woostack-eval/scripts/validate.mjs
howarewoo: Fixed in 5a60f1cf853ff1fbc4db8ef82fadf13d02c65848.

skills/woostack-init/references/legacy-migration.md
howarewoo: Fixed in 5a60f1cf853ff1fbc4db8ef82fadf13d02c65848.

skills/woostack-eval/scripts/validate.mjs
```diff
@@ -496,7 +506,69 @@ function validateUniqueStringArray(value, key, baseField, sourcePath, errors, va
...
+}
+
+function collectFixtureClassificationOracles(assertions) {
+  const values = new Set();
+  if (!Array.isArray(assertions)) return values;
+  for (const assertion of assertions) {
+    if (
+      assertion?.kind === 'final-json-path-equals'
+      && typeof assertion.pointer === 'string'
+      && assertion.pointer.split('/').at(-1) === 'classification'
```
howarewoo: **[High] Detect asserted output oracles beyond `/classification`**

The new semantic check collects expected values only when the asserted pointer ends in `classification`. It therefore still accepts a worker fixture such as `{"status":"complete"}` when the case asserts `/status == complete`, even though `references/runner.md` prohibits scenario-level/equivalent output-oracle fields. I reproduced this against this exact validator: it exited 0 with `valid: true`. Generalize the semantic guard to asserted output fields/aliases (including status and reason codes), and add a regression whose expected value is not also the classification value; the current `status: ambiguous` test passes only because it duplicates the classification oracle.

skills/woostack-init/scripts/tests/test-legacy-migration-contract.sh
```diff
@@ -0,0 +1,197 @@
...
+    if writable != (mode == "eligible"):
+        raise SystemExit(f"{case_id}: write capability does not match deletion eligibility")
+
+historical = cases["classifies-completed-records-from-exact-merge-evidence"]
+historical_fixture = json.loads(
+    (init / "evals/fixtures/migration/historical-completed.json").read_text()
+)
+if (
+    historical_fixture.get("remoteObservations") != []
+    or historical_fixture.get("receiptSet", {}).get("outcome") != "verified-git-only"
```
howarewoo: **[High] Keep the migration contract green after removing the oracle**

The fix correctly removes `receiptSet.outcome` from `historical-completed.json`, but this contract still requires `verified-git-only`. In canonical Linux, the exact old tip passes `legacy migration contract: PASS`; the current hosted head exits 1 with `historical migration must use Git-only proof and include overnight records`. Update this assertion to prove the historical classification from raw merge/ancestry/recovery evidence rather than reintroducing the worker-visible outcome field.

howarewoo (commented):
Second exact-head re-review: the four original migration threads are materially improved, but the current head still has two blocking validation regressions.

skills/woostack-eval/scripts/validate.mjs
howarewoo: Fixed in 21a2b6d645e484d75b7da7391efcc59c6aade973: the validator now checks string-valued output oracles from every final JSON assertion, with distinct status and reason-code regressions.

skills/woostack-init/scripts/tests/test-legacy-migration-contract.sh
howarewoo: Fixed in 21a2b6d645e484d75b7da7391efcc59c6aade973: the contract now proves historical classification from merge, provenance, and Git recovery evidence without restoring the removed outcome oracle.

skills/woostack-eval/scripts/validate.mjs
```diff
@@ -496,7 +506,67 @@ function validateUniqueStringArray(value, key, baseField, sourcePath, errors, va
...
+    return null;
+  }
+  if (value === null || typeof value !== 'object') return null;
+  for (const [key, child] of Object.entries(value)) {
+    const childLocation = `${location}/${escapeJsonPointerToken(key)}`;
+    if (
+      isFixtureOutputOracleKey(key)
+      && !(key === 'expected' && location === '' && isExpectedActionIdentityList(child))
+    ) return childLocation;
+    if (typeof child === 'string' && outputOracles.has(child)) return childLocation;
```
howarewoo: **[High] Preserve legitimate input values when scanning output oracles**

`outputOracles.has(child)` treats every string-valued expected result that also appears anywhere in a fixture as answer leakage. That rejects legitimate passthrough inputs: this repository’s own `woostack-eval/evals/fixtures/proposal-state.json` contains the approval digest that the case must report, so `bash skills/woostack-eval/scripts/tests/test-critical-corpora.sh` now fails at this exact head with `corpus-fixture-output-oracle` at `/proposalSnapshot/digest`. Limit this check to semantic answer fields/pointer-equivalent aliases (or provide an explicit way to mark legitimate observed inputs), and keep a package-wide regression proving every shipped critical corpus validates.

howarewoo (commented):
CHANGES REQUESTED — exact-head validation found one blocking regression. `test-critical-corpora.sh` fails at this head; GitHub reports no checks.

skills/woostack-eval/scripts/validate.mjs
howarewoo: [High] Avoid both global string false positives and non-string oracle gaps. This exact-value scan rejects legitimate raw inputs whenever an asserted string must be echoed/identified: the shipped woostack-eval package now fails validation because `/proposalSnapshot/digest` equals the asserted digest; woostack-init also fails on raw `/legacyRecords/0/phase`. At the same time it ignores booleans/numbers/null/arrays/objects: a minimal fixture `{"status":true}` with `/status == true` validates successfully. Consequently `test-validate.sh` passes while the production critical-corpus contract fails (`skills/woostack-eval package validation failed`). Model the output boundary by asserted pointer/key plus explicit semantic aliases and JSON-deep expected values, while allowing raw evidence fields required to derive/echo receipts; add regression cases using the shipped packages and non-string outputs.

skills/woostack-init/scripts/tests/test-legacy-migration-contract.sh
```diff
@@ -0,0 +1,220 @@
...
+)
+if (
+    historical_fixture.get("remoteObservations") != []
+    or any(record.get("phase") != "complete" for record in historical_fixture["legacyRecords"])
+    or merge_evidence.get("repositoryId") != historical_fixture.get("repositoryId")
+    or not isinstance(merge_evidence.get("pullRequest"), int)
+    or not isinstance(merge_evidence.get("mergeCommit"), str)
+    or not merge_evidence["mergeCommit"]
+    or set(merge_evidence.get("coversPaths", [])) != historical_paths
+    or merge_evidence.get("verified") is not True
```
howarewoo: [High] Prove merge/ancestry from raw Git/PR evidence instead of accepting a derived `verified` flag. The worker-visible historical fixture still contains `mergeEvidence.verified: true`, `provenanceLedger.verified: true`, `gitRecovery.*Verified: true`, and `phase: complete`, directly revealing the asserted verification/status booleans. This test checks those summaries and a non-empty merge SHA, but supplies no commit graph/PR observation from which exact merge ancestry can be independently derived. The updated contract now exits 0, yet the defective semantic shortcut remains. Replace the derived flags with raw bounded merge/ancestry observations and make the test compute/verify the relationship; ensure the fixture validator rejects derived boolean aliases.

howarewoo (commented):
Latest-head re-review found two remaining deterministic blockers in the validator/migration fixes.

skills/woostack-eval/scripts/validate.mjs
howarewoo: Fixed in `7db3fae4a1ffec8e321ab040d5178a2f42edb360`.

howarewoo: Fixed in `7db3fae4a1ffec8e321ab040d5178a2f42edb360`.

skills/woostack-init/scripts/tests/test-legacy-migration-contract.sh
howarewoo: Fixed in `7db3fae4a1ffec8e321ab040d5178a2f42edb360`.

skills/woostack-eval/scripts/validate.mjs
```diff
@@ -496,7 +511,102 @@ function validateUniqueStringArray(value, key, baseField, sourcePath, errors, va
...
+  for (const assertion of assertions) {
+    if (
+      assertion?.kind !== 'final-json-path-equals'
+      || !isJsonPointer(assertion.pointer)
+      || assertion.pointer === ''
+    ) continue;
+    const tokens = assertion.pointer.slice(1).split('/').map(decodeJsonPointerToken);
+    const key = normalizeOracleKey(tokens.at(-1));
+    const aliases = SEMANTIC_FIXTURE_ORACLE_KEYS.get(key);
+    const derivedBoolean = typeof assertion.expected === 'boolean' && key.endsWith('verified');
```
howarewoo: [High] Derive the private-output boundary from every asserted output path, not only this short semantic map and Boolean keys ending in `verified`. At this exact head I created tracked synthetic packages whose worker-visible fixture exposed the exact private `final-json-path-equals` value under its asserted key. `/phase` (string), `/blocked` (Boolean), `/currentHeadId` (null), `/writesAttempted` (array), and `/result` (object) all validated with exit 0. A `/reasonCode` control was correctly rejected, while legitimate raw `/proposalSnapshot/digest` input remained accepted. So a corpus can still hand the worker the answer for arbitrary output fields and pass validation. Please add a regression matrix that rejects an exact assertion-key/path match for every JSON value type while retaining legitimate values such as fixture `phase` used to derive a distinct output `status`.

skills/woostack-init/scripts/tests/test-legacy-migration-contract.sh
```diff
@@ -0,0 +1,305 @@
...
+    return found
+
+
+merge_commit = pull_request.get("mergeCommit")
+head_commit = pull_request.get("headCommit")
+if (
+    derived_fields(historical_fixture)
+    or pull_request.get("repositoryId") != historical_fixture.get("repositoryId")
+    or pull_request.get("state") != "MERGED"
+    or not isinstance(pull_request.get("number"), int)
```
howarewoo: [High] Require a positive, non-Boolean canonical PR number before treating the observation as exact merge evidence. `isinstance(value, int)` accepts negative integers (and Python Booleans). I changed the bounded fixture's PR number to `-1`, updated only the private expected provenance to `pr:-1@<merge-prefix>`, and this exact contract still printed `test-legacy-migration-contract: ok`. That leaves the previously requested invalid-PR mutation accepted even though the new SHA/ancestry checks correctly reject the old fake merge object. Use a strict positive-integer check and pin both negative and Boolean regression cases.

howarewoo (commented):
Exact-head rereview after the latest followups. The production corpus now validates all 20 packages without the earlier legitimate-input false positives, and the migration fixture now exposes raw Git/PR observations instead of self-certifying booleans. Two residual adversarial gaps remain below.

skills/woostack-eval/scripts/validate.mjs
howarewoo: Clarification from a second exact-head probe: this should not become a global same-value rejection—the `proposalSnapshot.digest` passthrough control must remain valid. The residual gap is that the semantic boundary registry recognizes only three families plus `*Verified`. Unambiguously worker-produced fields still passed for every value class: `rendered: false`, `selectedRound: 3`, `nextAction: null`, `writesAttempted: []`, and `result: {decision: "go"}` all matched their same-named private final assertion and validated with rc=0. Please expand the semantic output-field/alias contract (or introduce an explicit raw-observation schema) rather than comparing values globally; keep digest and phase→status passthrough controls to prevent the old false positives.

skills/woostack-init/scripts/tests/test-legacy-migration-contract.sh
howarewoo: I also ran the Boolean mutation rather than relying only on Python type semantics: with `pullRequestObservation.number: true` and the private expected provenance changed to `pr:True@<merge-prefix>`, the unmodified exact-head contract again printed `test-legacy-migration-contract: ok`. Both negative and Boolean cases are currently reproducible.

howarewoo: Additional independent exact-head counterfactuals confirm the same migration-proof boundary is broader than PR-number canonicality. Each of these isolated fixture mutations also left the unmodified contract green: `receiptComplete: true`, a duplicate `changedPaths` entry, a duplicate recovery observation, replacing the non-provenance base commit with the reserved all-zero 40-hex OID, and a self-cycle in the commit graph. All five printed `test-legacy-migration-contract: ok`. Please reject derived-summary aliases, duplicates before set/map normalization, the null OID, and cyclic/non-DAG observations in the same raw-evidence contract.

skills/woostack-eval/scripts/validate.mjs
howarewoo: One further independently reproduced bypass: `collectFixtureOutputBoundaries()` only inspects `final-json-path-equals`. With worker-visible `{scenario: "ambiguous"}` and private `/classification == "ambiguous"`, otherwise-valid tracked packages using `json-path-equals` and `receipt-field-equals` both validated with rc=0/codes=none. The semantic boundary needs to cover every assertion kind carrying a private expected output, especially worker-controlled file JSON, while retaining the paired raw-passthrough controls.

howarewoo: Fixed in 17a42a64d2ca2c49782473efca2db1ac91306277.

skills/woostack-init/scripts/tests/test-legacy-migration-contract.sh
howarewoo: Fixed in 17a42a64d2ca2c49782473efca2db1ac91306277.

skills/woostack-init/scripts/tests/test-legacy-migration-contract.sh
```diff
@@ -0,0 +1,398 @@
...
+        observation["path"]: observation
+        for observation in recovery_entries
+    }
+    number = pull_request.get("number")
+    head_commit = pull_request.get("headCommit")
+    merge_commit = pull_request.get("mergeCommit")
+    if (
+        derived_fields(value)
+        or len(historical_paths) != len(historical_path_list)
+        or pull_request.get("repositoryId") != value.get("repositoryId")
```
howarewoo: **[High] Bind migration proof to the canonical repository and reject derived-status aliases, rather than accepting internally consistent fixture claims.** At exact hosted head `17a42a64d2ca2c49782473efca2db1ac91306277`, I copied the historical fixture into separate writable Linux trees, applied one mutation per tree, and reran `test-legacy-migration-contract.sh`. Each of these still exits 0 with `test-legacy-migration-contract: ok`: (1) change both repository IDs to `repo-foreign`; (2) change both repository IDs to Boolean `true`; (3) add worker-visible `receiptCompleted: true`. Line 163 only compares the two fixture-authored IDs, so a foreign or type-invalid identity authenticates itself; `derived_fields()` recognizes suffix `complete` but not the equivalent `completed` alias. This conflicts with `legacy-migration.md`'s requirement to verify the canonical repository and block a foreign repository, and still lets a worker-visible derived completion summary act as proof. Please bind the observation to bounded canonical repository context with strict type/value validation, detect equivalent derived-status aliases, and add counterfactual regressions that fail on the defective behavior.

skills/woostack-eval/scripts/validate.mjs
```diff
@@ -44,6 +49,29 @@ const BEHAVIOR_CASE_FIELDS = new Set([
...
+  'expected',
+  'expectedAnswer',
+  'expectedOutput',
+  'expectedResult',
+  'oracle',
+  'outputOracle',
+  'terminalReceipt',
+]);
+const SEMANTIC_FIXTURE_ORACLE_KEYS = new Map([
+  ['classification', new Set(['classification', 'label', 'outcome', 'scenario'])],
```
howarewoo: **[High] Derive this boundary from every private asserted output path; the expanded finite semantic map still allows arbitrary output keys.** At exact head `17a42a64d2ca2c49782473efca2db1ac91306277`, I generated tracked packages whose assertion targets `/verdict` and whose worker-visible fixture contains the same `verdict` value. I exercised string, Boolean, number, null, array, and object values across `final-json-path-equals`, `json-path-equals`, and `receipt-field-equals`. All 18 packages were accepted (`rc=0`, no diagnostics). `/verdict` is not in `SEMANTIC_FIXTURE_ORACLE_KEYS`, so `collectFixtureOutputBoundaries()` discards the assertion even though the private assertion itself defines the output boundary. The shipped raw-evidence control (`proposalSnapshot.digest` → `/requiredApprovalDigest`) and all 20 production packages remain valid, so this should not become a global same-value ban. Please make arbitrary asserted output paths participate structurally, preserve legitimate differently-scoped raw evidence, and add generic-key regressions for every JSON value type and all three assertion surfaces.

skills/woostack-init/scripts/tests/test-legacy-migration-contract.sh
howarewoo: Fixed in 0b077e4a016d138da118c2b7623e544a62161f34.

skills/woostack-eval/scripts/validate.mjs
howarewoo: Fixed in 0b077e4a016d138da118c2b7623e544a62161f34.

skills/woostack-init/scripts/tests/test-legacy-migration-contract.sh
```diff
@@ -0,0 +1,410 @@
...
+        raise SystemExit("historical migration record provenance is not exact")
+    prefix, path = provenance[0].removeprefix("git:").split(":", 1)
+    matching_commits = [commit for commit in commits if commit.startswith(prefix)]
+    recovery = recovery_observations.get(record["path"], {})
+    recovery_file = (
+        init
+        / "evals/fixtures/migration/historical-completed/git-recovery"
+        / record["path"]
+    )
+    digest = "sha256:" + hashlib.sha256(recovery_file.read_bytes()).hexdigest()
```
howarewoo: **[High] Bind historical recovery bytes to the recorded Git object instead of a standalone fixture file.** At exact head `0b077e4a016d138da118c2b7623e544a62161f34`, `test-legacy-migration-contract.sh` still accepts migration evidence that cannot prove the historical files are recoverable from `sourceCommit:path`:

- I replaced both `project/` and `git-recovery/` files with arbitrary new bytes and updated only the fixture-authored SHA fields/private file-SHA assertions. The recorded `sourceCommit` and commit graph were unchanged. Result: `test-legacy-migration-contract: ok`.
- I separately moved both provenance/recovery `sourceCommit` values to the merge commit's base-side lineage (not the observed PR-head lineage) and updated only the private preserved-provenance expectation. Result: `test-legacy-migration-contract: ok`.

The current check hashes `git-recovery/<path>` directly and tests that the chosen commit is somewhere under the merge commit, but never derives or verifies that those bytes are the blob at the recorded `sourceCommit:path`. That leaves the proof self-authenticating: standalone bytes, fixture hashes, and fixture graph can agree while the recorded Git object does not contain the historical file. This conflicts with the approved migration contract requiring immutable Git blob/PR provenance and verification that every historical file is recoverable from the recorded Git object. Please bind recovery to an independently observed blob identity/content for the exact `sourceCommit:path` (and the canonical PR lineage), with counterfactual tests where standalone recovery bytes/hashes are rewritten consistently but the Git object is not. Reproduction summary: `source-outside-pr-head rc=0`; `rewritten-standalone-bytes rc=0`; `ACCEPTED_UNBOUND_MIGRATION_PROOFS=2`.

skills/woostack-eval/scripts/validate.mjs
```diff
@@ -496,7 +524,101 @@ function validateUniqueStringArray(value, key, baseField, sourcePath, errors, va
...
+}
+
+function decodeJsonPointerToken(value) {
+  return value.replaceAll('~1', '/').replaceAll('~0', '~');
+}
+
+function collectFixtureOutputBoundaries(assertions) {
+  const boundaries = [];
+  if (!Array.isArray(assertions)) return boundaries;
+  for (const assertion of assertions) {
```
howarewoo, line 571: **[High] Preserve legitimate same-key raw observations; the arbitrary-key fix currently invalidates half the production corpus.** The new `?? new Set([key])` fallback correctly rejects the 18 `/verdict` oracle probes, but it also treats every fixture field whose normalized key/value matches a private assertion as an oracle, even when that field is the raw observation the worker must inspect or echo. Canonical Linux package-wide validation at exact head `0b077e4a016d138da118c2b7623e544a62161f34` reports `ALL_EVALS packages=20 invalid=10 errors=31`; the immediate parent reports `packages=18 invalid=0 errors=0`, and the prior #571 head also passed all 20 packages. Examples now rejected include raw `/officialMcp/endpoint`, `/gitRead/worktrees/0/branch`, `/scenarios/0/events/0/data/issueId`, `/scenarios/1/issue/role`, `/project/events/10/data/operationId`, and `/safeCreate/clientId`. These are native/raw identities and observations, not worker-visible private verdicts. This is the precision half of the original requirement: reject semantically output-only fields for arbitrary assertion paths without globally rejecting legitimate raw evidence merely because its key/value is also asserted. Please add an explicit structural distinction for declared raw observations/passthroughs (or redesign those fixture boundaries), keep all 18 arbitrary `/verdict` cases red, and require package-wide production validation to remain green.

skills/woostack-init/scripts/tests/test-legacy-migration-contract.sh
howarewoo: Additional exact-head confirmation using real hosted OIDs: I set the fixture's source/head commit to parent `e9f94cde842ecd9a78c11fd6dcb08e8e19278895` and merge commit to current head `0b077e4a016d138da118c2b7623e544a62161f34`, updated the fixture graph/provenance consistently, and left the standalone recovery bytes in place. Both recorded paths are absent from the claimed source commit (`git cat-file -e <source>:<path>` exits `128` for each), yet the migration contract still prints `test-legacy-migration-contract: ok`. Together with the rewritten-bytes case, the current matrix is `ACCEPTED_UNBOUND_MIGRATION_PROOFS=3`. This isolates the missing `sourceCommit:path` Git-object binding from merely malformed or synthetic SHA concerns.

howarewoo: One more independently confirmed raw-observation mismatch at the same exact head: live GitHub reports canonical repository node ID `R_kgDORP1kWw`; PR #418 head `caf1bbead6f95d50436d47f84951df120b6f2c58`, merge `6d83ed6bc146e703ca66bdaf8076306d0a260b76`, and changed paths `.woostack/plans/2026-06-19-seo-angle-refine.md`, `skills/woostack-review/prompts/angles/aeo.md`, and `skills/woostack-review/prompts/angles/seo.md`. The committed fixture instead self-declares repository `repo-woostack`, head `3f90ca1f…`, merge `71e4d2f1…`, and two `cache-race.md` paths. Since the migration test accepts those claims as exact evidence without binding to bounded raw PR/Git observations, it also authenticates a PR identity/changed-path set directly contradicted by GitHub.

howarewoo: Fixed in 514e03cf70c5d16d286af4f22f66fa5d02225639.

skills/woostack-eval/scripts/validate.mjs
howarewoo: Fixed in 514e03cf70c5d16d286af4f22f66fa5d02225639.

skills/woostack-eval/scripts/validate.mjs
```diff
@@ -496,7 +524,111 @@ function validateUniqueStringArray(value, key, baseField, sourcePath, errors, va
...
+      let suffix = '';
+      for (let index = locationKeys.length - 1; index >= 0; index -= 1) {
+        suffix = locationKeys[index] + suffix;
+        if (suffix === boundary.key) {
+          matches = true;
+          break;
+        }
+      }
+    }
+    return matches && !boundary.passthrough;
```
howarewoo: **[High] Scope passthrough declassification to a concrete raw fixture location.** `fixturePassthroughAssertions` currently marks the whole assertion boundary as passthrough (`matchesFixtureOutputBoundary` returns `matches && !boundary.passthrough`), so one declaration suppresses every matching semantic key/value in every fixture for that case—not just the legitimate raw observation it is meant to preserve. At exact head `514e03cf70c5d16d286af4f22f66fa5d02225639`, I built 18 cases covering all six JSON value types and all three private assertion kinds. Each case had a legitimate raw fixture with the value at `/observations/verdict` plus a second worker-visible fixture leaking the same expected value at the output-shaped root path `/verdict`; adding `fixturePassthroughAssertions:["verdict"]` made all 18 packages validate: `DECLARED_SCOPE_BYPASS_VALID=true CASES=18 ERRORS=0`. As a paired control, removing only the declarations made the same corpus invalid with 36 oracle errors: `UNDECLARED_ORACLE_VALID=false ERRORS=36`. This means a declaration for one raw field masks an unrelated private output oracle elsewhere in the case. Please bind declassification to an exact raw source (for example assertion ID + fixture path + JSON pointer/raw namespace), suppress only that location, and reject unused/mismatched declarations. The production scan is back to 20/20 valid, so the precision repair itself is working; the blocker is the declaration's global scope.

skills/woostack-init/scripts/tests/test-legacy-migration-contract.sh
```diff
@@ -0,0 +1,414 @@
...
+if validate_historical_recovery(wrong_source_fixture, recovery_root) is not None:
+    raise SystemExit("historical migration accepts recovery from outside the canonical PR head")
+
+preserved_provenance = next(
+    assertion["expected"]
+    for assertion in historical["assertions"]
+    if assertion["id"] == "migration-historical-provenance"
+)
+if (
+    set(recovery_observations) != historical_paths
```
howarewoo: **[High] Restore historical migration coverage for `.woostack/overnight/`.** The Git-object repair replaces the prior historical `fix` + `overnight-handback` records with one real `plan` record and removes the contract guard that required an overnight fixture. A current corpus inventory reports `NO_OVERNIGHT_MIGRATION_FIXTURE`: none of the migration cases now exercises an overnight path. That weakens a load-bearing migration boundary rather than only strengthening its evidence. The spec explicitly detects `.woostack/specs/`, `plans/`, `fixes/`, and `overnight/` records, and `legacy-migration.md` requires the source ledger to cover every specification, plan, fix, and overnight path exactly once. The previous test also explicitly failed unless the historical fixture included `.woostack/overnight/`. Please retain the new real Git-object binding while adding representative overnight historical recovery/provenance coverage (using another bounded real observation or a deterministic disposable Git graph/blob fixture). Otherwise the suite can pass while the historical classification/deletion path for overnight handbacks is completely unproved.

skills/woostack-eval/scripts/validate.mjs
```diff
@@ -496,7 +524,111 @@ function validateUniqueStringArray(value, key, baseField, sourcePath, errors, va
...
+
+function collectFixtureOutputBoundaries(assertions, passthroughAssertions) {
+  const boundaries = [];
+  if (!Array.isArray(assertions)) return boundaries;
+  for (const assertion of assertions) {
+    if (
+      !PRIVATE_EXPECTED_ASSERTION_KINDS.has(assertion?.kind)
+      || !Object.hasOwn(assertion, 'expected')
+      || !isJsonPointer(assertion.pointer)
+      || assertion.pointer === ''
```
howarewoo: **[High] Enforce oracle isolation for whole-document assertions.** `collectFixtureOutputBoundaries` skips every private expected-value assertion whose RFC 6901 pointer is the empty string, even though `""` is valid here and means the complete final/file/receipt JSON value. At exact head `514e03cf70c5d16d286af4f22f66fa5d02225639`, a canonical-Linux matrix placed each private whole-document expected value under an arbitrary worker-visible fixture key (`payload`). All 18 combinations—six JSON value types across `final-json-path-equals`, `json-path-equals`, and `receipt-field-equals`—were accepted without any passthrough declaration: `WHOLE_DOCUMENT_PRIVATE_ORACLE_VALID=true CASES=18 ERRORS=0`. This leaves a complete-output oracle path outside the new fallback. Please derive a whole-document boundary as well (with exact-location raw declassification where needed), or reject empty-pointer private assertions when worker-visible fixtures are present; the current unconditional `assertion.pointer === ''` skip cannot preserve the required all-value/all-private-assertion isolation.

skills/woostack-init/scripts/tests/test-legacy-migration-contract.sh
```diff
@@ -0,0 +1,414 @@
...
+        return None
+    return historical_paths, pull_request, recovery_observations
+
+
+def validate_historical_recovery(value, recovery_root):
+    validated = validate_historical_raw_evidence(value)
+    if validated is None:
+        return None
+    historical_paths, pull_request, recovery_observations = validated
+    if not git_commit_exists(EXPECTED_HEAD_COMMIT) or not git_commit_exists(EXPECTED_MERGE_COMMIT):
```
howarewoo: **[High] Make the Git-object migration proof hermetic.** The new contract binds to PR #418's head `caf1bbe…`, but that PR was squash-merged: the reachable merge commit is `6d83ed6…` with only `b1855de…` as its parent, so `caf1bbe…` is not reachable from `main` or this PR. I created a complete bundle containing only exact head `514e03cf…` and its reachable history, cloned it in canonical Linux, and confirmed `pr418_head_present=no`, `squash_merge_present=yes`; `test-legacy-migration-contract.sh` then failed with `historical migration must bind canonical PR evidence to the recorded Git object` (`migration_rc=1`). It passes only after separately fetching the dangling `refs/pull/418/head`. A fresh/offline clone or normal CI fetch therefore cannot run the owning contract, and the external pull ref may eventually disappear. Please make the test self-contained—e.g. construct a deterministic disposable Git graph/blob plus bounded observation fixture, or ship a bounded fixture object store—while retaining the real `sourceCommit:path` byte lookup and counterfactual rejection. The test must not require a network fetch of an unrelated historical PR head.

skills/woostack-eval/scripts/validate.mjs
howarewoo: Fixed at exact hosted head `76348e367a4fe66185fc71037304b044d2a4df09`. Passthrough declarations are now exact objects binding `assertionId`, fixture path, and JSON pointer. The validator scopes consumption to that fixture location and rejects unused or mismatched declarations. The isolated oracle matrix covers all six JSON value types across all three private assertion kinds, rejects a second leaked location, and accepts only the declared raw location; `test-oracle-isolation: ok`.

skills/woostack-init/scripts/tests/test-legacy-migration-contract.sh
howarewoo: Fixed at exact hosted head `76348e367a4fe66185fc71037304b044d2a4df09`. Historical migration now includes both canonical plan and `.woostack/overnight/` records, requires the exact `{plan, overnight}` kind set, and verifies each recovery object and provenance token. `test-legacy-migration-contract: ok` in isolated offline Debian/Node 22.

skills/woostack-eval/scripts/validate.mjs
howarewoo: Fixed at exact hosted head `76348e367a4fe66185fc71037304b044d2a4df09`. Empty RFC 6901 pointers now create whole-document boundaries instead of being skipped. Whole-document passthrough requires an exact assertion/fixture/root-pointer declaration; undeclared values and assertion-ID transfer are rejected across all six JSON value types and three private assertion kinds. `test-oracle-isolation: ok`.

skills/woostack-init/scripts/tests/test-legacy-migration-contract.sh
howarewoo: Fixed at exact hosted head `76348e367a4fe66185fc71037304b044d2a4df09`. The migration contract now constructs a deterministic disposable Git repository with base, historical head, and merge commits, then verifies `sourceCommit:path` bytes, topology, digests, counterfactuals, and recovery without fetching any pull ref or using network access. `test-legacy-migration-contract: ok` in isolated offline Debian/Node 22.

howarewoo:
### Merge activity

* **Jul 31, 2:06 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/howarewoo/woostack/571).
* **Jul 31, 2:19 AM UTC**: [Graphite](https://app.graphite.com/github/pr/howarewoo/woostack/571) rebased this pull request as part of a merge.
* **Jul 31, 2:19 AM UTC**: @howarewoo merged this pull request with [Graphite](https://app.graphite.com/github/pr/howarewoo/woostack/571).

